PR #2, opened 2015-12-06
Make Linux installation more straight forward

teran-mckinney:
This could probably be smaller and simpler, but it's a start.

User will have to wait for lbrycrd to finish syncing before they
can do anything. I guess that should be a part of lbrynet-console.

### DIFF
--- a/view/page/linux.php
+++ b/view/page/linux.php
@@ -7,8 +7,7 @@
     <div class="span6">
       <h3>For the Efficient and Lazy</h3>
       <ol>
-        <li>Make a folder called <code>lbry</code> where you want everything to reside.</li>
-        <li>Download and run <a href="https://raw.githubusercontent.com/lbryio/lbry-setup/master/lbry_setup.sh" class="link-primary">this script</a> from that folder.</li>
+        <li><code>(mkdir lbry; cd lbry; wget https://raw.githubusercontent.com/lbryio/lbry-setup/master/lbry_setup.sh; chmod 755 lbry_setup.sh; ./lbry_setup.sh; bin/lbry/lbrycrdd -server -daemon)</code></li>
       </ol>
     </div>
     <div class="span6">


view/page/linux.php
```diff
@@ -7,8 +7,7 @@
     <div class="span6">
       <h3>For the Efficient and Lazy</h3>
       <ol>
-        <li>Make a folder called <code>lbry</code> where you want everything to reside.</li>
-        <li>Download and run <a href="https://raw.githubusercontent.com/lbryio/lbry-setup/master/lbry_setup.sh" class="link-primary">this script</a> from that folder.</li>
+        <li><code>(mkdir lbry; cd lbry; wget https://raw.githubusercontent.com/lbryio/lbry-setup/master/lbry_setup.sh; chmod 755 lbry_setup.sh; ./lbry_setup.sh; bin/lbry/lbrycrdd -server -daemon)</code></li>
```
jimmykiselak: This should all be good except for the last command. I don't think we want the installer to also run lbrycrdd. lbrynet-console and lbrynet-gui launch lbrycrdd when they start up and, if they launched it, shut it down when they exit. That way it's only running if the user wants it to be running. (Also, the path to the binary will be bin/lbrycrdd.) Thoughts?

teran-mckinney: Ahh, I didn't realize that lbrynet-console launched lbrycrdd. Does it kill it on exit? Seems like something that would be good to keep running, or at least have the option of such.

I agree with what you're saying in that case.

teran-mckinney: I see you updated the site. Unfortunately, the copy/pasting pastes it without newlines. You may want to add the semicolons back or find a way to put in the newlines without it taking them out.

I think for this a different text field might be ideal. Something like what http://brew.sh does.

kauffj: @jimmykiselak slightly confused by your comment: doesn't the last command do what you want (run the console)?

@teran-mckinney Can you clarify on copy/pasting without new lines? It pasted fine for me both directly to console and to GEdit on Ubuntu. I may fiddle with styles more as well.

kauffj: Ohhhh these comments were on a pull request, not on my last commit. We were literally doing this at the same time, lol.

jimmykiselak:
>Does it kill it on exit? Seems like something that would be good to keep running, or at least have the option of such.

It kills it if and only if it started it. If lbrycrdd was already running when lbrynet-console was launched, lbrynet-console will not kill it. If lbrycrdd is not running when lbrynet-console is launched, then lbrynet-console will start it and then stop it when it exits.

PR #10, opened 2016-07-01
add faq entries around credits and hosting

jobevers:


jobevers:
Had to make a fork because I don't have write access to the repo.

PR #19, opened 2016-07-08
Added dev links

idontcar3:
Added Development links (slack & github) to the navigation bars


kauffj:
@idontcar3 thanks for this fix, I made this change but made it as a separate commit due to other required changes. Check 133de62.

PR #45, opened 2016-07-14
blog footer, refactor to view partials, remove old content

kauffj:
Being a good dev and PRing rather than merging myself. Adds a footer to all blog posts to drive joining waitlist and social follows. @jobevers or @lyoshenka can you review?


controller/Controller.class.php
```diff
@@ -75,7 +75,10 @@ public static function execute($uri)
         // THIS ROUTE IS DEPRECATED. IT WILL BE REMOVED SOON.
         return static::redirect('https://github.com/lbryio/lbry/releases/download/v0.2.5/lbry_0.2.5_amd64.deb', 307);
```
jobevers: I know this isn't related to the PR, but the deprecation notice on this caught my eye.

jobevers:
I skimmed through it and it seems fine.  Is there a way to preview it - either by running locally or do branches get staged somewhere?

kauffj:
Branches aren't staged (this would be a good idea @lyoshenka). You can run it locally by running `./dev.sh` if you have PHP installed. As PHP dependencies grow this will get more annoying.

controller/Controller.class.php
kauffj: @jobevers updated deprecated endpoints to point to `/get`

PR #46, opened 2016-07-14
Referral fixes

kauffj:
@lyoshenka not sure how this got so FUBAR'ed, should only be  a28df97


jobevers:
when I do this, which is more then I'd like to admit, I usually cherry-pick the one commit onto a new branch:

```
git checkout master
git pull --rebase
git checkout -b referral_fixes_2
git cherry-pick a28df97
git push
```

kauffj:
@jobevers good idea and done, thanks!

PR #51, opened 2016-07-21
I18n

brunoparga:
Add basic internationalization framework and first version of Portuguese translations.


kauffj:
@brunoparga this PR contains several places where the files are mismerged, including the homepage (CTRL+F <<<<).

After a complicated merge it is good to do this search. And you should also test the common pages manually before submitting.

PR #59, opened 2016-07-27
Adding claimtrie implementation FAQ document

kaykurokawa:
First draft, should not be merged yet. Waiting for reviews and comments.


kauffj:
Really excited about this! I suspect work like this causes one to end up with a credit on the LBRY white paper Zargham is working on.

Would love to see what @jimmykiselak @lyoshenka and/or @jobevers think about this.

PR #70, opened 2016-09-02
Update earn-credits.md

emersonprocessmanagement:


kauffj:
Thanks, if you're in our Slack, DM kauffj and say hello!

PR #71, opened 2016-09-02
Spelling fix

shojo1:


kauffj:
Thanks! If you're in our Slack, DM kauffj and say hello!

PR #73, opened 2016-09-03
Internationalization of lbry.io

loblao:
As on https://lbry.io/bounty/web-i18n

This pull request implements a way to let the user pick a language.

Important notes:
- I skipped "Reverse look-up your IP and choose a default language if a user has not selected one" because i18n will deduce it from the host/subdomain so I thought this would be rather redundant. Furthermore, I didn't want to relay on a thirdparty for this. However, if this is really essential, this is trivial to implement.

- The front-end parts are pretty much a prototype (I'm not exactly good at design). The dropdown is really ugly and needs work but I believe it's outside the scope of this bounty. Again, if it's really necessary, I could come up with something. However, I'd rather leave this to a professional web designer of LBRY's.

I'll be on Slack to discuss this.


kauffj:
Took a look at this. Great work! And as you suspected, you don't need to worry about design. Some feedback:

- When you open your PR, please have it be off current version of master.
- Please add one canonical place that gets all available cultures, e.g. `i18n::getAllCultures()`. Currently available cultures are hard-coded in both `_globalItems` and `setCulture`.
- Small: it's probably not worth doing an AJAX POST just to redirect on success. Why not do a real post and then 301 to the currect URL? Also, this way the endpoint will not return an empty success if called via non-AJAX.
- Small: can we call endpoint `/set-culture` or something other than just `/language`?
- Minuscule: since we're on PHP7 (as of a few days ago), you can now do: `$_POST['culture'] ?? null`.

Additionally, I do think the geolookup is worth adding. If the user just comes into www.lbry.io, it will default to English when it may make sense to default to something else.

Well done and very close to what we were looking for!

PR #74, opened 2016-09-03
Fixed then/them typo

cariadeccleston:


kauffj:
Thanks! If you're in our Slack DM me at kauffj and say hello!

PR #75, opened 2016-09-03
fixing typo

edprince:
Small grammatical fix


kauffj:
Thanks! If you're in our Slack DM kauffj, happy to tip!

PR #76, opened 2016-09-11
I18n

kauffj:
@lyoshenka this is good to go, except:

- Language switcher should be hidden until we switch a flag to turn on i18n
- `i18nActions::setCulture` should redirect to the appropriate sub domain


lyoshenka:
merged

lyoshenka:
I did not enable the i18n subdomains. I'm not sure that's the best way to go

PR #80, opened 2016-10-22
session flashs, form styles, basic dmca form

lyoshenka:
NOTE TO SELF: install php7.0-bcmath before merging


kauffj:
:clap: not only does it fix the core problem, several auxiliary things cleaned up or added as well! Great work. It did not actually send the email for me, but I assume that is related to your personal note.

PR #87, opened 2016-12-06
new bounties and rev rewards

kauffj:
Mainly needs review to decide how to handle the request for bindings. I would also revise the READMEs on sample repo before doing this.

### DIFF
--- a/content/bounty/app-i18n.md
+++ b/content/bounty/app-i18n.md
@@ -1,0 +1,18 @@
+---
+category: browser
+title: Internationalization of LBRY Browser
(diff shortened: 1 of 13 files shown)


lyoshenka:
I'm gonna merge this in but remove the bindings bounty for now. I think it probably should be separate repos, and we should figure out how it works before we start asking people for contributions. For example, I'd like a minimum standard for the code (tests, documentation, error handling, etc).

If you want to flesh this out some more, I'd merge it in. I don't have time for it right now.

PR #94, opened 2017-03-18
Updated with the correct place to find the wallet

filipnyquist:
Updates/change the locations from the old non existing default_wallet in lbryum Main dir to wallets/default_wallet.


gapracoolz:
i agree...its make easy to find the file

PR #95, opened 2017-03-22
Quickstart: add detail to metadata fields

androm3da:


view/template/developer/_quickstartCredits.php
```diff
@@ -19,7 +19,16 @@
   "name": "electricsheep",
   "file_path": "\\home\kauffj\\Desktop\\electric-sheep.mp4",
   "bid": 1,
-  "metadata":  { } <span class="code-bash__comment">//this should match the metadata returned by resolve_name </span>
+  "metadata":  {
+    "description": "Some interesting content",
+    "title": "my_interesting_content",
```
crasm: Perhaps titles should be mainly formatted for humans? Like "My Interesting Content" or "My interesting content".

view/template/developer/_quickstartCredits.php
```diff
@@ -19,7 +19,16 @@
...
+    "author": "Video shot by me@example.com",
+    "language": "en",
```
crasm: For language, would there be any restriction on what kind of values should be entered? For instance [ISO-639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). Otherwise there would be content with `"language": "en" ` and `"language": "English"`, etc, which would make filtering more difficult.

Also, what about some sort of null or not-applicable value? In my case, a photograph of a bird doesn't have a language.

view/template/developer/_quickstartCredits.php
```diff
@@ -19,7 +19,16 @@
...
+    "license": "LBRY Inc",
+    "ver": "0.1",
```
nikooo777: this cannot be specified AFAIK

view/template/developer/_quickstartCredits.php
```diff
@@ -19,7 +19,16 @@
...
+    "nsfw": false,
+    "sources": {}
```
nikooo777: this is not required either AFAIK

crasm: According to the schema it's still required, but maybe it's filled in automatically when publishing files?

androm3da:
Tried to address all comments, except regarding `language`.  I agree w/your comments, @crasm just not sure what the resolution should be.

For the most part it's probably not critical that it be addressed in the "quickstart" but what kind of `string` is acceptable here should be documented somewhere.

crasm:
Looks good to me, and I agree about the language property being tricky. I just used "none" as my language string for now.

kauffj:
Thank you @androm3da and @crasm!

Going to make a few other tweaks here as well.

PR #98, opened 2017-04-01
file_path in publish example consistency

flungo:
The `file_path` in the publish example is now consistent with a path that would be used in a UNIX based system rather than some weird inconsistent Windows/Unix hybrid. Just an aesthetic change, really.


kauffj:
Thank you!

PR #102, opened 2017-04-25
Update _quickstartApi.php (#1)

bsinglet:
Updated the Quickstart guide to API version 0.10.0, which requires a "uri" parameter for GET requests instead of a "name" parameter.


kauffj:
Wonderful, thank you @bsinglet!

bsinglet:
You're welcome!

PR #109, opened 2017-05-24
add wolfram chart

tzarebczan:
Thought it would be a good idea to reference. I'll try to re-create the chart n another tool with date references as opposed to just Days since Oct 2016


lyoshenka:
I'd rather not link to wolfram here. If you want to make a nice-looking image of the block subsidy graph, I would include that.

PR #122, opened 2017-06-16
add footnote to credit withdrawal

tzarebczan:
Address credit withdrawal on main page


kauffj:
Good call, thanks @tzarebczan

PR #136, opened 2017-07-05
Added instuctions for changing port.

hackrush01:
@lyoshenka 
Goes with https://github.com/lbryio/lbry/pull/749. Is this the only file that needs to be added?


Makgolli:
How do you change the port on Linux? I have no conf.py under .lbrynet.

tzarebczan:
Please see https://lbry.io/faq/how-to-change-port

PR #139, opened 2017-07-11
Provide an introduction on using LBRY API using python.

hackrush01:
Basic introduction is given along with 2 examples which should be enough for a developer to understand the usage.


lyoshenka:
I think this needs a bit more in the beginning. The first example starts with 

```from lbrynet.daemon import get_client```

How do you get `lbrynet`? Do you `pip install` it?

lyoshenka:
You should also explain that the daemon needs to be running, that RPC will happen over the default port, etc.

hackrush01:
Closed due to some problem(Thank you very much git...) Updates in #191

PR #140, opened 2017-07-12
Two FAQ entries: lbrynet..log and clean install

tzarebczan:
Created two FAQ entries:
1) How to Find lbrynet.log for each OS
2) How to clean install and preserve wallet/application data

### DIFF
--- a/content/faq/how-to-clean-install-preserve
+++ b/content/faq/how-to-clean-install-preserve
@@ -1,0 +1,36 @@
+---
+title: How to do a clean install of LBRY while preserving important data
+category: setup
+---
+
+Under certain circumstances, it may be recommended that you perform a clean installation of LBRY.  This generally involves uninstalling the current LBRY application, removing various directories/files and then reinstalling the LBRY application.  Since each Operating System has its own set of working directories, it's a good idea to review: https://lbry.io/faq/lbry-directories before proceeding.  
+
+After uninstalling LBRY, the working directories are left intact so that the application can be easily reinstalled without losing your application data and wallet file  Application data is in the lbrynet folder (LBRY folder for Mac) and it includes various database and configuration files.  Wallet and blockchain data is in the lbryum folder. Typically, you would want to keep your wallet and a handful of configuration/database files when performing a clean install while keeping most of your important data intact.
+
+## Windows
+1. Uninstall the LBRY application by going to "Add or Remove Programs" via the Control Panel
+2. Type in %appdata% into an Explorer window to find the working directories
+3. Delete the entire LBRY folder
+4. Navigate to the lbryum folder and delete the blockchain_headers file
+5. Navigate to the lbrynet folder and delete all items while leaving the following:
+5.1 'blobfiles' folder - stores blobs of downloaded files for hosting purposes
+5.2 'blobs.db' - reference data for the blob files which are used for hosting purposes
+5.3 'lbryfile_info.db' - Downloads and Publishes data
+5.4 'install_id' - Links your LBRY account with email address. Delete if want to start with new email. 
+- `C:\Program Files (x86)\lbrynet` - LBRY application itself
+- `C:\Users\%USER%\AppData\Roaming\lbrynet` - Daemon configuration and blobs
+- `C:\Users\%USER%\AppData\Roaming\LBRYum` - Wallet and blockchain headers (if using the app or the lbryum wallet)
+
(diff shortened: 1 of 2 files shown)


content/faq/how-to-clean-install-preserve
```diff
@@ -0,0 +1,57 @@
+---
+title: How do I do a clean install of LBRY while preserving important data?
+category: setup
+---
+
+Under certain circumstances, particularly if you've tested earlier beta versions, it may be recommended that you perform a clean installation of LBRY.  This generally involves uninstalling the current LBRY application, removing various directories/files and then reinstalling the latest version of LBRY. Different operating systems use different folder locations and [this post on directories](https://lbry.io/faq/lbry-directories) will help if you cannot locate the right files or folders.
```
kauffj: Added some light clean up to this file. Most importantly, please always link email addresses and URLs, and typically URLs should be entered with a label rather than entered directly.

content/faq/how-to-clean-install-preserve
```diff
@@ -0,0 +1,57 @@
...
+2. Uninstall the LBRY application by accessing "Add or Remove Programs" via the Control Panel
+3. Type `%appdata%` into an Explorer window to find the working directories
+4. Delete the entire LBRY folder
+5. If performing a clean install, delete the lbryum and lbrynet folders also **(!!THIS WILL DELETE YOUR WALLET AND DATA!!)** and skip to step 8 
+6. Navigate to the lbryum folder and delete the `blockchain_headers` file
+7. Navigate to the lbrynet folder and delete all items while leaving the following:
+ 1. `blobfiles folder` - stores encrypted downloaded files which are used for hosting
+ 2. `blobs.db` - reference data for the blob files which are used for hosting purposes
+ 3. `lbryfile_info.db` - Downloads and Publishes data
+ 4. `blockchainname.db` - Supports downloads data
```
kauffj: I removed the install_id reference.

content/faq/how-to-clean-install-preserve
```diff
@@ -0,0 +1,57 @@
...
+8. Install the latest verison of LBRY from: [Github App Page](https://github.com/lbryio/lbry-app/releases "Github App Page"). If prompted to allow through Windows Firewall, click Allow
+9. LBRY should start immediately after install.  If you kept your data, your balance and content will be reflected
+
+## MacOS
+1. Reboot your PC (this is to ensure that no LBRY processes are running)
+2. Uninstall the LBRY application by dragging the LBRY app from the Applications folder to the Trash (located at the end of the Dock), then choose Finder > Empty Trash
+3. Open Finder, click Go on top menu, choose "Go To Folder", Type "~/Library/Application Support/LBRY" and then click go.
+4. If performing a clean install, delete the entire contents of this folder **(!!THIS WILL DELETE YOUR LBRY DATA!!)** and proceed to step 6
+5. Otherwise, delete everything except for:
+ 1. `blobfiles folder` - stores encrypted downloaded files which are used for hosting
```
kauffj: Could this article be restructured so that more of the common parts are shared?

content/faq/how-to-clean-install-preserve
tzarebczan: Thanks for the edit, looks good.

content/faq/how-to-clean-install-preserve
tzarebczan: Got it. I would assume a new install id would be generated and they would be re-prompted for their email/ verification code - is that what we want? If the code is now expired, they will need one re-sent.

content/faq/how-to-clean-install-preserve
tzarebczan: Let me take a look to see what can be restructured, but I think each OS is different enough that a majority of it is specific to each section.  Maybe the part about keeping files can and reinstalling can be made into a separate section, but the files reside in different directories for Mac vs Linux (LBRY vs lbrynet).

content/faq/how-to-clean-install-preserve
kauffj: The codes don't expire, but we will likely want to add an additional send of the code to this step.

However, we actually switched from using the install ID to using operating system level keyring storage. This should persist regardless.

lyoshenka: email verification codes do expire. if they enter a correct but expired code, they'll be automatically send a new valid one

PR #182, opened 2017-07-14
fix channel name, add additional info

tzarebczan:
Changed #bugs to #help
Re-wording
Add lbry-app repo link


kauffj:
What about help@lbry.io?

tzarebczan:
Added, good call!

PR #187, opened 2017-07-27
Update app-i18n.md

Cryptolized:
It's "its"; not "it's" ;)


kauffj:
Thanks for these @Cryptolized!

PR #189, opened 2017-07-27
Update slack-lbry-url-handler.md

Cryptolized:
I assume that is what it should say.


tzarebczan:
Bounties sent, thanks!

PR #191, opened 2017-07-31
Provide an introduction on using LBRY API using python.

hackrush01:
Updated with better instructions.

Replaces #139.


kauffj:
I don't see a need for a FAQ article specifically for Python, unless we're intending to add these for all languages. 

Additionally, the recommended way to interact with the daemon is just to use the JSONRPC server, the same any other language would use.

PR #207, opened 2017-08-19
Make Index Of Bounties Page's

MSFTserver:
Create an Index of Bounties So They Are Easy To Find


kauffj:
This exists. See https://lbry.io/bounty

PR #213, opened 2017-08-24
how to publish guide

dpmidd:


content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
+## How to publish content on LBRY
+
+Advanced users: All of the steps below are for the LBRY app.  For additional details of how to conduct any of these actions through the cli please see https://lbryio.github.io/lbry/cli/
```
tzarebczan: Can you give a little intro to what publishing is in LBRY

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+
+### How to create a channel
+1. Download the LBRY app from lbry.io/get. If you already have the application and it prompts you to update, run the update to be on the most recent version.
```
tzarebczan: The download link should be in the intro step.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+2. Load the application and  click the `Publish` button.
+3. In the `Channel Name` section click the dropdown and select `New Channel` and  declare the name you would like for your channel.  For more details on different channel types, see https://lbry.io/faq/naming
+4. Once your name is selected, there is a `Deposit` section that is below.  It requires a minimum bid of .1 LBC.   See more on deposits here: naming.md.  If you would ever like to abandon your claim there is a guide below in `Abandoning a claim`.
```
tzarebczan: We don't really have a minimum or a way to calculate one right now (it's like a TX fee, it varies) - but if you had to give one, I'd say 0.0001 LBC to be safe. 

"See more on deposits here: naming.md". - naming faq url?

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+5. Click `Create Channel` once you have selected your bid amount.  Your deposit is a bid and if this channel isn't claimed by another bid, the channel is yours to use!
```
tzarebczan: "Your deposit is a bid and if this channel isn't claimed by another bid, the channel is yours to use!" - either remove or explain that you now own lbry://@channelname#<channel> and lbry://@channelname if you are the highest bidder.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+You can revoke this claim if you no longer want to use it with the steps below.
```
tzarebczan: see the previous comment, abandoning channels is only available via the CLI. Also, it's risky right now because you have to make sure all claims from the channel are abandoned first, and if not, you break your wallet.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Publish Content to a Channel
+
+1. Under the `Content` section click `Choose File`.
+2. Select what you would like to upload to LBRY.  LBRY accepts any HTML5 format for video, the full list can be found here: https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats
```
tzarebczan: Maybe reference that HTML formats are for streaming video, but other file types can be downloaded and viewed externally. I think this section will also get a revamp down the road to be more explicit in the video types.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+3. Select a `Title`, `Thumbnail URl`, and `Description` for your upload.
```
tzarebczan: `Thumbnail URl` - URL 

Maybe expand on that and it clear that this needs to be a hyperlink to an image file or to use spee.ch

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+4.  Below those there is a `Language` and `Maturity` which will default to `English` and `All Ages`.  If a change is needed to these defaults click the dropdowns and change them.
+Under the `Access` section select what type of license the content should be under, select which type of license is appropriate for your upload. </br>
```
tzarebczan: select, select again - wording
Can you find anything online that might explain these copyright options?

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+5. Select the channel name you would like to publish the channel under.
```
tzarebczan: This would be an optional step - otherwise, it gets posted as "Anonymous" and will not be part of a channel.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+6. Type out the full url you want the content to be found under along with a minimum of .1 LBC minimum deposit for the upload.
```
tzarebczan: probably a good place to reference https://lbry.io/faq/naming again

See my previous note about min fee, same applies to content bid

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+7. Read and agree to the terms of service.
+8. Click publish.
```
tzarebczan: Tell them the file will process in the background and will be added to the LBRY Blockchain. We don't have a way to monitor status at this time, but mention that larger files may take a couple minutes to process and not to shut down until the file is out of "pending confirmation" mode.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Abandoning a claim and deleting uploaded content
+
+1. Click on the folder icon in the top right.
+2. Click on the `Published` link.
+3. Click on the `...` next to the `Open` link.
```
tzarebczan: missing a step - click on the content you want to remove, then the ...

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+4. Select `Remove...`.
+5. There will be two options.  `Abandon the claim for this URI` and `Delete this file from my computer`.  Select the option that applies.  <b>Warning: Deleting content is permanent.  Please make sure this is what you want to do before confirming the deletion.
```
tzarebczan: you can bold something by putting ** ** around the text. See markdown syntax 

Mention that abandoning the URI releases the original bid back into your wallet.  Warning would be that if it's not abandoned, the content may still exist on the LBRY network.  Also, if it's removed without abandoning, it still shows in Publishes -https://github.com/lbryio/lbry-app/issues/493 

Maybe by default we should recommend the abandoning, I really don't see the use case otherwise.

content/faq/how-to-publish.md
```diff
@@ -0,0 +1,39 @@
...
+6. Click `Remove`.
+
+### Editing your claim on a channel
+
+1. Go to the `Publish` page.
+2. In the `Content URL` section type in the route of the claim of the content you would like to edit.
+3. Below the URL you will see text saying "You already have a claim with this name." and a link that says `Use data from my existing claim`.  Click the link.
+4. You can now edit your uploaded claim information.
```
tzarebczan: When they are done, hit re-confirm TOS, hit publish.  Right now there is a bug that makes them reselect the file, but hopefully, this is fixed soon.

content/faq/how-to-publish.md
dpmidd: I decided not to include anything on linking to abandoning claims since the naming faq is already referenced, it will be in my commit that I'm pushing soon.

lyoshenka:
@tzarebczan im closing this. reopen if you get back to it.

PR #214, opened 2017-08-31
add YouYube Sync and manual verification

tzarebczan:
add links for youtube sync / slack
add note about CC verification being the fastest

Feel free to give feedback/add/edit.


kauffj:
1) I thought YT sync was instant?
2) I think directly linking to the sync page is confusing. Maybe this is a separate FAQ resource? I would say use lbry.io/youtube , but it kinda sucks.

PR #247, opened 2017-10-31
Add route to clear cache

lyoshenka:
Going to `/clear-cache` will clear the APC cache

### DIFF
--- a/controller/Controller.class.php
+++ b/controller/Controller.class.php
@@ -2,6 +2,8 @@
 
 class Controller
 {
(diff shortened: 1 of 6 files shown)


kauffj:
@lyoshenka two potential changes:

- Is it an issue that anyone can clear the APC cache? Especially since this is an open-source project...
- Can the post-commit hook that updates the website hit this on every deploy?

lyoshenka:
1. not worried about that, but its easy to lock down if you want
2. done

PR #248, opened 2017-11-07
Create lbry-rewards

LavRadis:
FAQ on LBRY rewards as discussed with Tom.


tzarebczan:
Thanks for the feedback @kauffj, I've been collaborating with Lavris on this and will take a closer look.

The support link is lbry.io/faq/how-to-report-bugs (I'll add a section for email help here)

kauffj:
@tzarebczan I'd propose renaming that article to just "support" as well (i.e. lbry.io/faq/support)

If you do this, add a redirect here so the old link works: https://github.com/lbryio/lbry.io/blob/master/controller/Controller.class.php#L128

PR #283, opened 2017-12-07
create youtube sync FAQ

tzarebczan:
Please provide feedback / make edits as needed


wilsonk:
Hey @tzarebczan, in the 'Minimum requirements' paragraph, the second sentence should probably reference the number of views/videos again to keep things clear.  So 'If it does not, please use one of the other' should probably be 'If your channel does not meet these requirements, please use...'.

PR #284, opened 2017-12-08
reilly edits

reillysmith:
I want the rewards out of YouTube sync. It has no place here and confuses the purpose of YouTube sync to the reader, which is to save a publisher labor if they have hundreds of archival videos. YouTube Sync is not about rewards, and we need to stop making it part of rewards verification.

### DIFF
--- a/content/faq/youtube.md
+++ b/content/faq/youtube.md
@@ -3,12 +3,4 @@ title: What is YouTube Sync?
 category: publisher
 order: 4
 ---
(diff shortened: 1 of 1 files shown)


tzarebczan:
Per our discussion today, closing this. @reillysmith can you make any other suggestions given the rewards path is stayin?

reillysmith:
I have nothing else to contribute here.

PR #286, opened 2017-12-09
Update mining-credits.md

ykris45:
adding note. What do you think ? And please fix wording if interest. Thanks


ykris45:
Hooray merged again thanks @tzarebczan. Try make a pr in chrome android .

PR #287, opened 2017-12-09
Update mining-credits.md

ykris45:
Fixing character symbol , before " \ " thats is false in the directories structure. Now "/" , thanks
PS : please review after merged because i am using chrome android in my smartphone for male this PR. Thanks


tzarebczan:
It's an escape character to account for spaces. It's correct.

PR #305, opened 2017-12-31
added lbrypool.md to list

painbank:
just trying to raise some visibility.  Would love to include calls to the apis of the different pools and list the current fees here as well, but I'm not sure where all to add that code yet.


tzarebczan:
Thanks @painbank!

PR #309, opened 2018-01-08
add two missing dependencies

kauffj:
Locally, the blog did not work without installing `php-xml` and `php7.0-mbstring` via aptitude.

This is an attempt to add them via composer, but I was unsure how to test/verify that this is correct (`php composer.phar install` runs fine, but does that mean this is correct?).


lyoshenka:
this won't actually install the dependencies, but at least it will error if you don't have them installed

PR #322, opened 2018-01-17
add lbry-app-basics

FUNKePILLs:
FAQ for beginners to the LBRY app.


content/faq/lbry-app-basics
```diff
@@ -0,0 +1,127 @@
+---
+title: LBRY App Basic User Guide
```
dpmidd: I think this is more of a debugging FAQ than a basic user guide. Thoughts? Basic user guides to me would outline functionality and describe what is where, UI/UX, etc

content/faq/lbry-app-basics
```diff
@@ -0,0 +1,127 @@
...
+category: LBRY 101
+---
+
+Installing LBRY and setting it up can be a bit tough, especially if you encounter bugs or errors.
```
dpmidd: Shouldn't start the document with saying installing is tough and then have the first topic be download/stream failing.  Maybe something more generic like, "Using LBRY is easy but sometimes you can run into bugs, here is a list of commonly asked questions and solutions."

kauffj:
@FUNKePILLs thanks for this! There is some good info in here, but we really want each LBRY FAQ entry to stand on it's own as a single question, rather than a single comprehensive "User's Guide". @tzarebczan could possibly work with @FUNKePILLs to figure out which of this information is covered elsewhere, which merits it's own question, and which ought to be omitted entirely?

tzarebczan:
@kauffj I originally gave @FUNKePILLs some hints for this and then had other community members give it a look over before I did a final review.  The original goal was to introduce a few basic concepts of the LBRY app that a new user should be aware of.  

I agree that some of the information is repeated  - we'll need to fix that up.  I don't think it hurts mentioning these items, but a simple link back to the appropriate FAQ will suffice.

PR #325, opened 2018-01-19
Add twitter to team member bio

maximest-pierre:
fix #313 

Tested it locally and it works.

Could add Github id easily also.

PR #326, opened 2018-01-19
Add github id to team page

maximest-pierre:
As discussed in #325 add GitHub id for the team member.

I can only see 7 members in the organization page. Can someone confirm that everyone is there?
screenshot from 2018-01-19 10-29-27


kauffj:
Looks good @maximest-pierre! @tzarebczan will fill in all LBRY team member Twitter and GitHub details.

Issue #327, opened 2018-01-20
Incompatible with PHP 7.2


flungo:
Due to one of the vendored dependencies (`lib/vendor/scss`) the site is incompatible with PHP 7.2. When using 7.2, the following error occurs without which then results in the content not being rendered:

> <br />
> <b>Deprecated</b>:  The each() function is deprecated. This message will be suppressed on further calls in <b>/usr/src/lbry.io/lib/vendor/scss/Node/Number.class.php</b> on line <b>293</b><br />
> <pre>exception 'ErrorException' with message 'Cannot modify header information - headers already sent by (output started at /usr/src/lbry.io/lib/vendor/scss/Node/Number.class.php:293)' in /usr/src/lbry.io/view/Response.class.php:296
> #0 unknown file(unknown line): {closure}(2, 'Cannot modify header information - headers already sent by (output started at /usr/src/lbry.io/lib/vendor/scss/Node/Number.class.php:293)', '/usr/src/lbry.io/view/Response.class.php', 296, Array(3))
> #1 /usr/src/lbry.io/view/Response.class.php(296): header('Content-Security-Policy: frame-ancestors 'none'')
> #2 /usr/src/lbry.io/view/Response.class.php(136): Response::sendHeaders()
> #3 /usr/src/lbry.io/controller/Controller.class.php(52): Response::send()
> #4 /usr/src/lbry.io/web/index.php(34): Controller::dispatch('/get')
> </pre>

Obviously, this only breaks the site when developing (since setting `$config['is_prod'] = "yes";` disables reporting). Either the dependency needs fixing/updating such that it does not use the deprecated function or a more specific version range should be specified in the composer metadata (and possibly a note in the `README.md`) to make the supported versions clear.

Issue #297, opened 2017-12-21
Mac Docker Setup Broken


liamcardenas:
I'm using Docker version 17.09.1-ce on macOS Sierra 10.12.6. The `docker.sh` script doesn't work. This seemed to mostly work

```
#!/bin/bash

DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
scriptDir="$(cd "$(dirname "$0")" && pwd -P)"

echo $scriptDir


if [ ! -e "data/config.php" ]; then
  cp "$DIR/data/config.php.example" "$DIR/data/config.php"
fi

docker run --rm -it --name "dev.lbry.io" \
  -v "${scriptDir}:/usr/src/lbry.io" \
  -w "/usr/src/lbry.io" \
  -p "127.0.0.1:8000:8000" \
  -u "$(id -u):$(id -g)" \
  php:7-alpine \
  php --server "0.0.0.0:8000" --docroot "web/" "web/index.php"
```

But then it said that the directory `web/` could not be found. I am not sure how to fix this, I have instead opted to run with just php for now which works fine.

lyoshenka:
This script was added by an outside contributor. It probably needs to be updated or removed, since I don't think any of us use it.

kauffj:
@flungo are you interested to fix this?

flungo:
I'll take a look at the weekend make sure it still works on Linux and see if there is anything I can spot that would mean it doesn't work on Mac although I do not have a Mac to be able to test/confirm compatibility.On 16 Jan 2018 21:21, Jeremy Kauffman <notifications@github.com> wrote:@flungo are you interested to fix this?

flungo:
Made a patch (#328) which should be compatible with MacOS.

@liamcardenas could you please checkout my fork and confirm that this now works for you?

Bear in mind #327 which will mean you should set `$config['is_prod'] = "yes";` in `data/config.php` or change `php:7-alpine` to `php:7.1-alpine` in the `docker.sh` script to be able to see the site correctly until the deprecated function is removed or compatibility is specified as `>=7.0 <7.2` and I'll update my PR to reflect this by making the change described here.

liamcardenas:
Hi @flungo awesome that you are working on this, unfortunately, I no longer have MacOS. Do you need someone else to test? or are you just trying to verify that my exact problem is not reoccurring.

I have access to a Mac, but it is not the same one that originally had the error.

Let me know if there is anything I can do.

[...]

PR #328, opened 2018-01-20
Fixed Docker script and README

flungo:
Updated the Docker script to start the site correctly. The DIR variable was not being used in each place which it should have been used and this has allowed the removal of the `readlink -f` command which had compatibility issues with MacOS (#297).

The README did not have the change in name for the docker script reflected in all references so this has also been corrected.

Note, this currently requires that `$config['is_prod'] = "yes";` is set in `data/config.php` due to #327 and the fact I currently have the script set to use the latest PHP 7 version (at present 7.2.1). Should you decide to specify a smaller subset of compatible PHP versions then I can update this PR to pin the docker version at either 7.0 or 7.1 depending on your preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ You can also run the development server using docker:
 - Run `./docker.sh` from the project root
 - Access [localhost:8000](http://localhost:8000) in your browser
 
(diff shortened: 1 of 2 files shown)


kauffj:
Thank you! Also, look for an email from me shortly.

PR #333, opened 2018-01-24
Add warning in roadmap and bold a warning

maximest-pierre:
Added a warning for when the API key for Asana and Github is not set. This warning could have saved me some time last night.

Bonus: The note at the end is now ***Bold*** to make it more visible.

### DIFF
--- a/view/template/content/roadmap.php
+++ b/view/template/content/roadmap.php
@@ -6,6 +6,12 @@
   lbry.roadmap('#project-roadmap');
 <?php js_end() ?>
 <main>
+  <?php if (ConfigurationAction::checkGithubKey() == false && ConfigurationAction::checkIsProd() == false){
+    echo "<h1>Github Key is not set api call are limited to 60 call per hour</h1>";
+  }?>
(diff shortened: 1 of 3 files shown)


view/template/content/roadmap.php
```diff
@@ -6,6 +6,12 @@
   lbry.roadmap('#project-roadmap');
 <?php js_end() ?>
 <main>
+  <?php if (Config::get('github_key') == null || Config::get('is_prod') === "no"){
```
kauffj: 1) Standards wise, this check should probably be in controller.
2) The `'is_prod'` check seems ugly / mistake prone. I checked `is_prod` setting on production web server and it is `"yes"` as a string, which _also_ seems ugly. @lyoshenka can you coordinate with @maximest-pierre on lbry.io access? This will let you see production config values. Likely `is_prod` should be boolean true, and checks for production should demand `=== true`.

maximest-pierre: I am implementing the check inside of a controller right now. I am opening an issue for the is_prod issue.

tzarebczan:
@kauffj did you review this one already? Can't tell what changes were requested

maximest-pierre:
I will close it and redo it later with the config changes since its going to affect it.

PR #335, opened 2018-01-26
fix coauthor to co-author

ykris45:
https://dictionary.cambridge.org/amp/english/co-author based on this site


tzarebczan:
coauthor is also a valid word

PR #339, opened 2018-01-29
minor fixes

ykris45:


data/i18n/en.yaml
```diff
@@ -236,7 +236,7 @@ social:
   humanstext: Join the LBRY community.
   robotsheader: Robots
   robotstext: Make with us. All LBRY code is open source.
-  slack: Slack (chat)
+  discord: Discord (chat)
```
kauffj: The left side of these declarations is the input to an i18n call, e.g. `__('social.slack'`)`. I think in this case this one isn't used, but generally you can't change the key declaration without also changing the i18n calls themselves.

ykris45: whether this can be changed. instead of now migrating to discord. sir

ykris45:
@kauffj can i change the key on https://github.com/lbryio/lbry.io/blob/0b4fa4368341772af32f40e78fd3cffcd4458510/controller/Controller.class.php line 87 to 'discord'  sir . hehe

tzarebczan:
@ykris45 that's something completely different. I'm closing this, thanks.

PR #340, opened 2018-01-29
Update _content.scss

ykris45:
twitterwidget showing red . it means error. i fixed with twitter-widget .


kauffj:
This is correct. See https://lbry.io/news/5-questions-about-lbry for a page that includes a `<twitterwidget>` element.

PR #352, opened 2018-02-01
Highlight links on roadmap

### LINKED ISSUE #347: highlight hyperlinks on roadmap
Continuation of https://github.com/lbryio/lbry.io/issues/292

The hyperlinks to prs/issues on https://lbry.io/roadmap are not highlighted as such

maximest-pierre:
fixes #347 highlight on roadmap links.

### DIFF
--- a/web/scss/_roadmap.scss
+++ b/web/scss/_roadmap.scss
@@ -118,7 +118,7 @@ $title-spur: 8 * $width-roadmap;
   font-size: 0.9em;
   margin: 10px;
 }
-
+.roadmap-item-content a{@include anchor($color-primary);}
 .roadmap-item-assignee
 {
   float: right;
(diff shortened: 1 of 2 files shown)


web/scss/_roadmap.scss
```diff
@@ -118,7 +118,7 @@ $title-spur: 8 * $width-roadmap;
   font-size: 0.9em;
   margin: 10px;
 }
-
+.roadmap-item-content a{color:$color-primary;}
```
kauffj: I think the proper solution to this is just to apply the `primary-link` tag to the anchor.

Even if a custom rule was preferred here, then it should presumably still use `@include anchor($color-primary)` as is done elsewhere.

kauffj: It'd also probably be acceptable to wrap the roadmap description items in the CSS class `markdown`.

maximest-pierre:
I have made the change proposed

PR #353, opened 2018-02-02
Fix/258 download button

Jeremy1026:
Created a new partial (download/_button) to manage a download button that automatically selects the proper OS. Applied this partial on the home page and on /get. If no download is available for the current OS the button on the home page links to /get which allows users to enter their email address to receive notifications of new releases.


controller/action/DownloadActions.class.php
```diff
@@ -28,6 +28,29 @@ public static function executeGetDaemonRedirect(string $os)
...
+  }
+
+  public static function prepareButtonPartial() {
+    if (static::prepareDownloadInformation()) {
+      return static::prepareDownloadInformation();
+    }
+    return [];
+  }
+
+  private static function prepareDownloadInformation() {
```
kauffj: `protected` over `private` generally. But in the case of `prepareXXX` functions, they should be `public`.

`prepareFoo` will be called automatically when including `_foo.php` as a view.

Jeremy1026:
Changes made per request @kauffj

kauffj:
@Jeremy1026 can you rebase on master? Looks like PR is bringing in unrelated commits.

tzarebczan:
@Jeremy1026 you may want to open a new PR for this unless you can re-base just your changes somehow. Looks like this one got pretty messy. Thanks!

tzarebczan:
@Jeremy1026 not sure if you tried to fix something (got a github notification) but it still looks messy.

tzarebczan:
@Jeremy1026 - closing this, but please take another shot when you are available :)

PR #354, opened 2018-02-02
Refactor config with constant instead of strings

### LINKED ISSUE #336: The production variable should be a Boolean and not a String
Currently, the is_prod configuration variable is a string. This means that the server can be is_prod ```'yes'``` or ```'no'```. This is ugly, we should change it for a Boolean to make it cleaner and easier to debug.

This variable is defined in the ```config.php```.

There are also checks here and there that also needs to be changed.

Steps to accomplish this:
1. Change the variable in the ```config.php``` and ```config.php.example```
2. Change all of the checks (The search function is your best friend in this case)
3. Profit and also help us make an awesome website

maximest-pierre:
This is a refactoring of the way we interact with the config file.
Fixes #336 

We can now access the configuration with constant instead of writing the string by hand each time.

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -8,7 +8,7 @@ if (php_sapi_name() === 'cli-server' && is_file(__DIR__.preg_replace('#(\?.*)$#'
 
 include __DIR__ . '/../bootstrap.php';
 
-define('IS_PRODUCTION', Config::get('is_prod') === "yes");
+define('IS_PRODUCTION', Config::get(Config::IS_PROD) === true);
 
 ini_set('display_errors', IS_PRODUCTION ? 'off' : 'on');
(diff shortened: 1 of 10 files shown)


web/index.php
```diff
@@ -8,7 +8,7 @@
 
 include __DIR__ . '/../bootstrap.php';
 
-define('IS_PRODUCTION', Config::get('is_prod') === "yes");
+define('IS_PRODUCTION', Config::get(Config::IS_PROD));
```
kauffj: Let's make this `=== true` in case other local setups still have values that are truthy (coerce to true via `==`).

maximest-pierre:
@kauffj Change has been made

PR #363, opened 2018-02-05
Add <?php to the example configuration.

### LINKED ISSUE #362: Example configuration doesn't work
The example configuration (`/data/config.php.example`) doesn't require properly, preventing the dev server from serving anything the first time, since it starts up with a copy of the example configuration. This is annoying since the warning is fairly late and disconnected from the source of the issue.

xorgy:
Resolves #362

### DIFF
--- a/data/config.php.example
+++ b/data/config.php.example
@@ -1,3 +1,5 @@
+<?php
+
 $config = [];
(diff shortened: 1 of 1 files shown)


lyoshenka:
Thanks for the PR. Check out https://lbry.io/faq/tips to claim a reward.

PR #366, opened 2018-02-05
added screenshot of how to delete

ykris45:
my initiatives


tzarebczan:
@ykris45 denying this because we can't have screenshots that show personal published files - it also does not look professional.

PR #367, opened 2018-02-05
move faq to header menu

ykris45:
this request so that visitors of this blog can go to link FAQ easily without click learn header menu


kauffj:
We're not going to do this. We can consider giving it more prominence on the /learn page.

PR #372, opened 2018-02-08
Google site verification for webmaster tools for Julie

jsigwart:
Needed access for managing sitemaps and structured data markup for site


kauffj:
@jsigwart we should already be verified. Please work with @finer9

kauffj:
https://github.com/lbryio/lbry.io/blob/master/web/google164701be10959176.html

PR #374, opened 2018-02-10
Open social links in new tab

jsigwart:
Open social media links in a new browser tab per Discord user suggestion.


tzarebczan:
Thanks Julie! Are there anywhere on the site, or thats it?

tzarebczan:
@jsigwart I updated the footers (left some links to open in the same window) for developers and Join us social links as well.

PR #384, opened 2018-02-19
Adding youtube onboarding phase 2

maximest-pierre:
Here what is missing before merging
- [x] Fix front-end error display


view/Response.class.php
```diff
@@ -106,11 +106,31 @@ public static function addJsAsset($src)
...
   {
     return static::$assets['js'];
   }
 
+  public static function getCssAssets()
+  {
+    return static::$assets['css'];
+  }
+
+  public static function setAllCss()
```
kauffj: I think this is the wrong approach. Rather than calling this everywhere, put `all.css` in the default asset definition (similar to JS). Then, add a method called `setCssAssets(array $assets = [])`.

When a page such as yt2 wants to disable use of all.css, it would call `Response::setCssAssets(['css/yt2.css'])`, which would remove the all default.

view/Response.class.php
```diff
@@ -106,11 +106,31 @@ public static function addJsAsset($src)
...
+  {
+    array_push(static::$assets['css'], '/css/all.css');
+  }
+
+  public static function setYT2Css()
```
kauffj: See above, this function should go.

view/template/acquisition/youtube_token.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
kauffj: Almost all of this code should be in the controller.

view/template/acquisition/yt2.php
```diff
@@ -213,5 +204,62 @@
 <script type="text/javascript" src="/js/yt2/TweenMax.min.js"></script>
 <script type="text/javascript" src="/js/yt2/ScrollToPlugin.min.js"></script>
 <script type="text/javascript" src="/js/yt2/app.js"></script>
-</body>
-</html>
+    <script language="javascript" type="text/javascript">
+        function submitDetailsForm() {
+            $("#youtube_claim").submit(function (event) {
```
kauffj: Put this code in a JS file and include it, rather than inlining all of it.

view/template/acquisition/yt2.php
```diff
@@ -66,20 +45,32 @@
...
+      <?php
+      // Change to something more meaningful
+      if (isset($_GET['error'])): echo "<div>error</div>";
+      endif;?>
     <div class="content">
       <div class="zigzag"></div>
       <h1>Own your identity. For real this time.</h1>
-      <form>
-        <div class="form-inner">
+      <form id="youtube_claim" method="post" action="/youtube/token/">
```
kauffj: This should be a 2 step form, not a form with 3 fields. The first step is just channel name. The second step is URL and email.

view/template/acquisition/yt2.php
```diff
@@ -66,20 +45,32 @@
...
+        <div class="form-inner" >
+            <div hidden id="lbry_error">LBRY channel name is not valid or blank</div>
```
kauffj: I was receiving this error for valid channels.

view/template/acquisition/yt2.php
```diff
@@ -66,20 +45,32 @@
...
-            <label>@</label>
+            <input id="lbry_channel_name" type="text" name="desired_lbry_channel_name" placeholder="Your Channel" />
           </div>
+            <div hidden id="email_error">Email is not valid or blank</div>
           <div class="block">
-            <input type="submit" value="Claim now" />
+            <input id="email" type="text" name="email" placeholder="Your Email"/>
           </div>
-        </div>
+            <div hidden id="youtube_error">Youtube channel url is not valid or blank</div>
```
kauffj: I was receiving this error for valid YouTube channel URLs.

view/template/acquisition/yt2.php
```diff
@@ -213,5 +204,62 @@
 <script type="text/javascript" src="/js/yt2/TweenMax.min.js"></script>
```
kauffj: Use `Response::addJsAsset`

view/template/acquisition/yt2.php
```diff
@@ -66,20 +45,32 @@
...
           </div>
-        </div>
+            <div hidden id="youtube_error">Youtube channel url is not valid or blank</div>
+          <div class="block">
+            <input id="youtube_channel" type="text" name="youtube_channel" placeholder="Your youtube channel url or channel id">
+          </div>
+
       </form>
+        <div class="block">
+            <input type="submit" value="Claim now" onClick="return submitDetailsForm()"/>
```
kauffj: Avoid inlining these. Use JS to attach them after.

view/template/acquisition/yt2.php
```diff
@@ -66,20 +45,32 @@
 </section>
 <section class="claim section">
```
kauffj: (This isn't the correct line number, but the reward values still need to be plugged in dynamically.)

controller/action/AcquisitionActions.class.php
```diff
@@ -42,13 +42,73 @@ public static function executeYouTube(string $campaignId = '')
...
+
+  public static function executeYoutubeStatus(string $token)
+  {
+    return ['acquisition/youtube_status', ['token' => $token]];
+  }
+
+  public static function actionYoutubeToken(string $email, string $desired_lbry_channel_name, string $youtube_channel_id){
+
+    $email_is_valid = static::email_verification($email);
+    $desired_lbry_channel_name_is_valid = static::lbry_channel_verification($desired_lbry_channel_name);
```
kauffj: Some inconsistencies here with casing and use of parantheses.

I don't have strong opinions about this, I just like consistency.

Perhaps there is a PHP equivalent of [PrettierJS](https://github.com/prettier/prettier) that could help with this.

controller/action/AcquisitionActions.class.php
```diff
@@ -42,13 +42,73 @@ public static function executeYouTube(string $campaignId = '')
...
+        Controller::redirect($url);
+      }
+    }
+    else{
+      $url = "/yt2?error=true";
+      Controller::redirect($url);
+    }
+  }
+
+  private static function email_verification($email){
```
kauffj: Unless you specifically need or want `private`, you should almost always use `protected` IMO.

view/template/acquisition/yt2.php
```diff
@@ -171,22 +169,22 @@
         </div>
         <div class="line">
           <p>1,000</p>
-          <p>250 <span></span></p>
+          <p><?php echo number_format($reward['data']['1000']); ?><span></span></p>
```
kauffj: Trailing semi-colons aren't necessary here

Issue #315, opened 2018-01-15
FAQ to summarize / simplify LBRY's stance and role in censoring "bad" content


tzarebczan:
Questions come up frequently about pirated, sensitive, illegal content and how LBRY plans to deal with it. 

The closest resource we have for this information is: https://lbry.io/what#combatting-the-ugly

A summarization/simplification of LBRY's stance and how it plans to combat that type of content should belong in a separate FAQ article.

tzarebczan:
Questions to consider covering:

Q. What happens if I use LBRY to publish kiddie porn?

A. https://lbry.io/what#combatting-the-ugly

Q. What happens if I use LBRY to publish the secret scripture of the Church of Scientology, and their lawyers want to sue both me and you into oblivion?

A. We comply with the CDA / DMCA to the best our technology allows. We would not be responsible for this, regardless of it's legality. Of course, given they're the church of scientology, nothing is probably going to actually stop them from suing us ;)

You, however, would be fully liable if you used LBRY to break the law.

Q. What happens if someone uses LBRY to publish my copyrighted material and I want to force it down and/or sue them into oblivion?

A. Similar to the linked answer above. 

Q. What happens if I use LBRY to publish classified material?

A. Again, similar answers to above. Not that dissimilar from what if someone did all of these things via BitTorrent.

tzarebczan:
@reillysmith would you want to take a stab at this one?

tzarebczan:
Maybe @jsigwart wants to take a shot after she's done with meetups/campus :p

jsigwart:
Yes, you can assign this one to me, @tzarebczan

tzarebczan:
@jsigwart can't assign you for some reason, but will create an Asana task to track this.

kauffj:
Was this finished?

tzarebczan:
Being worked on here: https://docs.google.com/document/d/1TynBaUgxfaufdVWjZqRrbKMr3zMY6sHexWwgNokl4wU/edit

jsigwart:
Going to add copy per Tom's request, "we should talk about the protocol/blockchain vs app in terms of censorship somewhere in this doc," but would love more eyes on the overall doc.

jsigwart:
#388 pull request

tzarebczan:
Closed in https://github.com/lbryio/lbry.io/pull/388

PR #388, opened 2018-02-21
Initial creation of Content FAQ re: #315

jsigwart:
Proposed text.

### DIFF
--- a/content/faq/content.md
+++ b/content/faq/content.md
@@ -1,0 +1,34 @@
+---
+title: What kind of content can I upload, and what about illegal or infringing content?
+category: publisher
+---
+
+This guide provides answers about questions regarding what you may upload to the LBRY app and network, as well as how to report illegal or infringing content.
+
+###What content can I legally upload to LBRY?
+
+You may upload content you created or own that does not infringe on the rights or safety of others, or content you have permission to upload. This includes Public Domain content, Creative Commons or Open Source content shared according to the license requirements, and Copyrighted content where you have permission from the owner.
+
+###Isn’t LBRY decentralized and the content can’t be removed?
+
+It is important to make a distinction between the LBRY blockchain/protocol and any applications running on top of them when referring to censorship. The LBRY protocol is fully decentralized and censorship resistant, with the protocol storing metadata and naming on the blockchain, and facilitating data transfers over a peer to peer (P2P) network. However, LBRY also makes an app to demonstrate our protocol. With our app we will engage in non-arbitrary censorship, meaning only horrific or infringing content will be removed. As a U.S. company LBRY Inc. and our app will follow all U.S. laws including DMCA. If someone made an app or website using the LBRY protocol in some other country, it would have to follow that country's laws, which aren’t necessarily the same as ours. Either app would read the same blockchain, though. 
+
+###Someone uploaded my content to LBRY without permission. How do I remove it?
+
+Reporting infringing/illegal content in the LBRY App is easy. Open the LBRY App and navigate to the content you wish to report. Underneath the content are two buttons: "Download," and "Report." When you click on the Report button, you will be redirected to a web form to report the content to LBRY, Inc., who can remove the content link from the LBRY App. 
+
+###What do I do if I see content that’s illegal or infringing in the LBRY App?
+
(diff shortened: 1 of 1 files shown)


content/faq/content.md
```diff
@@ -0,0 +1,34 @@
+---
+title: What kind of content can I upload, and what about illegal or infringing content?
+category: publisher
+---
+
+This guide provides answers about questions regarding what you may upload to the LBRY app and network, as well as how to report illegal or infringing content.
+
+###What content can I legally upload to LBRY?
+
+You may upload content you created or own that does not infringe on the rights or safety of others, or content you have permission to upload. This includes Public Domain content, Creative Commons or Open Source content shared according to the license requirements, and Copyrighted content where you have permission from the owner.
```
kauffj: Do we have the say "safety" here? Is that against the law?

(I'm all for discouraging people from uploading content that would make others unsafe, but the question specifically says "what content can I legally upload?")

content/faq/content.md
```diff
@@ -0,0 +1,34 @@
...
+
+###Isn’t LBRY decentralized and the content can’t be removed?
```
kauffj: This phrasing seems slightly awkward.

content/faq/content.md
```diff
@@ -0,0 +1,34 @@
...
+
+It is important to make a distinction between the LBRY blockchain/protocol and any applications running on top of them when referring to censorship. The LBRY protocol is fully decentralized and censorship resistant, with the protocol storing metadata and naming on the blockchain, and facilitating data transfers over a peer to peer (P2P) network. However, LBRY also makes an app to demonstrate our protocol. With our app we will engage in non-arbitrary censorship, meaning only horrific or infringing content will be removed. As a U.S. company LBRY Inc. and our app will follow all U.S. laws including DMCA. If someone made an app or website using the LBRY protocol in some other country, it would have to follow that country's laws, which aren’t necessarily the same as ours. Either app would read the same blockchain, though.
```
kauffj: - The LBRY protocol encompasses both the blockchain and data network. I don't think the slash is necessary.
- I'd paragraph break before the however. But I also have a preference for short paragraphs. 
- Mention CDA along with DMCA
- Definitely mention that LBRY makes no strong guarantees that data will be hosted. If no one is willing to host data, it won't continue to be available on the network.

content/faq/content.md
```diff
@@ -0,0 +1,34 @@
...
+
+###Someone uploaded my content to LBRY without permission. How do I remove it?
+
+Reporting infringing/illegal content in the LBRY App is easy. Open the LBRY App and navigate to the content you wish to report. Underneath the content are two buttons: "Download," and "Report." When you click on the Report button, you will be redirected to a web form to report the content to LBRY, Inc., who can remove the content link from the LBRY App.
```
kauffj: We can also just link the page here: https://lbry.io/dmca

content/faq/content.md
```diff
@@ -0,0 +1,34 @@
...
+
+Please follow the FAQ above for reporting infringing content, and LBRY staff will review your report and take any appropriate action. 
+
+###What is a DMCA request or takedown notice?
+
+The Digital Millennium Copyright Act (DMCA) is United States Copyright Law written to protect digital content against illegal use. When an owner finds their content used on the Internet without their permission, they can file a DMCA request with the website, app, or Internet Service Provider illegally hosting their content, asking for it to be removed.
+
+###How does LBRY respond to a DMCA request or takedown notice?
+
+LBRY Inc., will remove links deemed infringing from the LBRY App when it receives a valid DMCA request from the content owner.
```
kauffj: Probably more accurate to say something like "LBRY, Inc. will block access to content deemed infringing..."

Possibly superfluous to say "from the content owner". As I understand it, DMCA requests are _only_ valid if from the owner.

kauffj: Possibly worth mentioning that we provided an API that can allow you to check if a specific piece of content is infringing according to our records.

content/faq/content.md
jsigwart: I think by "safety" I was really referring to illegal content - like CP, threats, etc. Will modify.

kauffj:
@jsigwart thanks and merged! Only change I made was "peer to peer" -> "peer-to-peer"

PR #389, opened 2018-02-21
added shapeshift faq based on get credits menu LBRY App

ykris45:


tzarebczan:
Thanks @ykris45 , made some small changes and addition for bounty page.

PR #391, opened 2018-02-22
Create acceptable-use-policy.md

jsigwart:
Initial commit. #391

### DIFF
--- a/content/faq/acceptable-use-policy.md
+++ b/content/faq/acceptable-use-policy.md
@@ -1,0 +1,42 @@
+---
+title: Acceptable Use Policy
+category: publisher
+---
+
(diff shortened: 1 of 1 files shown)


tzarebczan:
Is this intended to be a publisher FAQ?

kauffj:
@tzarebczan this is a slow-moving reaction to lbrywallet.com and potentially similar situations.

jsigwart:
It was suggested this be done as a FAQ, and I added it as a Publisher FAQ. Thanks for catching that, Tom, because you're right, it is more broad than just publishing. Should we make a different category for AUP, Privacy Policy, and ToS? Trying to think of a good category name... suggestions?

tzarebczan:
@jsigwart we should put it in the Other category - I don't think we want to make a new category unless we have a few related items to go into it. If you don't agree, let's discuss.

jsigwart:
That makes complete sense. When we have more articles for a new category,
then we can create it and move this article.

tzarebczan:
@jsigwart should we merge this or it needs additional eyes?

jsigwart:
This is set to go now.

PR #395, opened 2018-02-23
Remove local version of jquery and add a cdn for it

### LINKED ISSUE #382: Update jquery
The version of jquery that is bundle is really outdated. The version bundle is 2.1.3 and the current one is 3.3.1.

An alternative to using a bundled library is to use a CDN to deliver the appropriate dependency.

maximest-pierre:
Fixes #382 

Switching from local jquery to CDN jquery.


ykris45:
it would be better to use local jquery. compared to using CDN, why I can say so. if the website code.jquery will not be accessible or in the hack, it will affect the stability of this website code. therefore I upload the file locally for better.

PR #396, opened 2018-02-23
Update twitter-tipbot.md

ykris45:
i fix the year of this bounty. before date is 2019 and now is  2018


tzarebczan:
Thanks Chris!

PR #406, opened 2018-03-01
Better Facebook Analytics

maximest-pierre:
Move the facebook analytic to the header like it says on the Facebook pixel instruction.

### DIFF
--- a/view/Response.class.php
+++ b/view/Response.class.php
@@ -32,7 +32,8 @@ class Response
     $contentSent = false,
     $isHeadersOnly = false,
     $gzipResponseContent = true,
-    $metaImages = [];
+    $metaImages = [],
+    $facebookAnalyticsType = "PageView";
 
   public static function setMetaDescription($description)
   {
@@ -350,6 +351,14 @@ class Response
     ];
 
     return $statusTexts[$code] ?? null;
+  }
+
+  public static function setFacebookPixelAnalyticsType($type){
+    static::$facebookAnalyticsType = $type;
+  }
+
+  public static function getFacebookPixelAnalyticsType(){
+    return static::$facebookAnalyticsType;
   }
 
   protected static function normalizeHeaderName($name): string
--- a/view/template/layout/_analytics.php
+++ b/view/template/layout/_analytics.php
@@ -16,18 +16,10 @@ h.end=i=function(){s.className=s.className.replace(RegExp(' ?'+y),'')};
     ga('require', 'GTM-NT8579P');
     ga('send', 'pageview');
 
-    <?php //we load fb sdk elsewhere, events loaded below (do we need both??) ?>
+   !function(f,b,e,v,n,t,s) {if(f.fbq)return;n=f.fbq=function(){n.callMethod? n.callMethod.apply(n,arguments):n.queue.push(arguments)}; if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0'; n.queue=[];t=b.createElement(e);t.async=!0; t.src=v;s=b.getElementsByTagName(e)[0]; s.parentNode.insertBefore(t,s)}(window,document,'script', 'https://connect.facebook.net/en_US/fbevents.js');
 
-    !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
-    n.callMethod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
-    n.push=n;n.loaded=!0;n.version='2.0';n.queue=[];t=b.createElement(e);t.async=!0;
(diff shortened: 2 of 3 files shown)


view/Response.class.php
```diff
@@ -352,6 +353,14 @@ public static function getStatusTextForCode($code)
     return $statusTexts[$code] ?? null;
   }
 
+  public static function setPixelAnalyticsType($type){
```
kauffj: Please put the word "Facebook" both in the protected variable name and function name.

Alternatively, this functionality could be dropped, since I think we're going to fire `PageView` on every page.

view/template/layout/_header_analytics.php
```diff
@@ -0,0 +1,10 @@
+<?php if (IS_PRODUCTION): ?>
```
kauffj: I don't see a strong reason to put this in `<head>`. Just leave it where it is, unless I'm misunderstanding and it doesn't work in the footer (it should).

view/template/page/college.php
```diff
@@ -1,6 +1,7 @@
 <?php Response::setMetaDescription(__('description.meetup')) ?>
 <?php Response::setMetaTitle(__('title.meetup')) ?>
 <?php echo View::render('nav/_header', ['isDark' => true, 'isAbsolute' => true]) ?>
+<?php Response::setPixelAnalyticsType("Lead") ?>
```
kauffj: Drop this, keep as PageView

view/template/page/meet.php
```diff
@@ -1,6 +1,7 @@
 <?php Response::setMetaDescription(__('description.meetup')) ?>
 <?php Response::setMetaTitle(__('title.meetup')) ?>
 <?php echo View::render('nav/_header', ['isDark' => true, 'isAbsolute' => true]) ?>
+<?php Response::setPixelAnalyticsType("Lead") ?>
```
kauffj: Drop this, keep as PageView

kauffj:
@maximest-pierre in addition to above we need to trigger an additional tracking event to `mail/_subscribeForm` when the user submits the form (yes, this will have a few false positives on invalid emails, but it's easiest solution, since they don't always end up on the same URL after submission).

This should look something like:
```
<?php js_start() ?>
  $('#form').submit(function() {
    fbq('track', 'Lead');
   //possibly other analytics events
  });
<?php js_end() ?>

**Comment by kauffj** [comment]
@finer9 do you want a GA event or Twitter event on email captures as well?

Note that as articulated above, we'll be firing the same event on ALL email captures, including those on pages like lbry.io/android. If that's not what is desired please clarify ASAP. If different events are required based on form, then we'd need to add an event parameter to pass to `_subscribeForm`.

File: view/template/mail/_subscribeForm.php [hunk 5]
```diff
@@ -1,7 +1,7 @@
 <?php $error = $error ?? null ?>
 <?php $tag = $tag ?? null ?>
 <?php $largeInput = $largeInput ?? false ?>
-<form action="/list/subscribe" method="POST" novalidate>
+<form id=mail_form action="/list/subscribe" method="POST" novalidate>
```
kauffj: Missing quotes.

view/template/layout/_analytics.php
```diff
@@ -16,18 +16,8 @@
...
-    n.callMethod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
-    n.push=n;n.loaded=!0;n.version='2.0';n.queue=[];t=b.createElement(e);t.async=!0;
-    t.src=v;s=b.getElementsByTagName(e)[0];s.parentNode.insertBefore(t,s)}(window,
-    document,'script','//connect.facebook.net/en_US/fbevents.js');
-
-    fbq('init', '1618717031725766');
-    fbq('track', 'PageView');
-
-    <?php //and now everyone knows what happens on our website except us ?>
+   !function(f,b,e,v,n,t,s) {if(f.fbq)return;n=f.fbq=function(){n.callMethod? n.callMethod.apply(n,arguments):n.queue.push(arguments)}; if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0'; n.queue=[];t=b.createElement(e);t.async=!0; t.src=v;s=b.getElementsByTagName(e)[0]; s.parentNode.insertBefore(t,s)}(window,document,'script', 'https://connect.facebook.net/en_US/fbevents.js');fbq('init', '1618717031725766');
```
kauffj: fix spacing or at least put the init on it's own line

view/Response.class.php
```diff
@@ -352,6 +352,14 @@ public static function getStatusTextForCode($code)
     return $statusTexts[$code] ?? null;
   }
 
+  public static function setFacebookPixelAnalyticsType($type){
+    static::$analyticType = $type;
```
kauffj: rename protected variable as well

finer9:
The same event name is fine because the FB side tracks them back to what campaign/adgroup and that is all that really matters.

PR #413, opened 2018-03-02
Add samuel bryan picture

maximest-pierre:
Because of the way the code is structured the best way to change the URL of the image is to change the name of the markdown file which required changing all author from lbry to samuel-bryan.


tzarebczan:
Makes sense, LGTM!

PR #414, opened 2018-03-05
fix wording

ykris45:
based on user in #general chat


nikooo777:
Tom is not around so I'll just merge this myself. Very straight forward typo.
Thanks

ykris45:
@nikooo777 thanks Niko... Tom on vacation right now

PR #429, opened 2018-03-15
Include info about the pools which donate to LBRY.Community

real-rouse:
:)


real-rouse:
Maybe remove the "<- Support the community with this pool" or is this all right?

tzarebczan:
@real-rouse removed the note at the bottom, think the arrows get the point across.

tzarebczan:
okay, i'll get it updated

PR #431, opened 2018-03-15
youtube sync faq 2.0

reillysmith:


tzarebczan:
@reillysmith add a section on how to redeem the rewards and what the current status of that is (not sure when they'll actually be able to redeem them). Maybe a good time to make a note of LBRY connected email > sync status page email.

PR #442, opened 2018-03-20
Update exchanges.md

hadi78m:
add trade by trade website


tzarebczan:
@hadi78m thanks so much for this! Looks like it doesn't allow US customers, so I can't really test it out. 

Please see https://lbry.io/tips if you'd like a tip for this contribution 💯

PR #458, opened 2018-03-21
Feature/yt2 sync update

### LINKED ISSUE #425: Convert yt2.css to SCSS
yt2.css should instead be an SCSS file called `youtube.scss`. It should be compiled the same way other SCSS files are.

maximest-pierre:
fixes #425  and add the final touch to youtube sync.


controller/action/AcquisitionActions.class.php
```diff
@@ -114,4 +114,14 @@ protected static function lbry_channel_verification($lbry_channel)
       return false;
     }
   }
+
+  public static function actionGetLBRYReward()
```
kauffj: You misunderstood the instruction/solution here.

It's:
```

  public static function executeYouTube()
  {
    return ['acquisition/youtube', [
      'reward' => LBRY::youtubeReward()
    ]];
  }
```

kauffj: please re-write all of these similar to above

view/View.class.php
```diff
@@ -139,8 +139,11 @@ public static function compileCss()
       $scssCompiler->setLineNumberStyle(Leafo\ScssPhp\Compiler::LINE_COMMENTS);
     }
 
-    $css = $scssCompiler->compile(file_get_contents(self::SCSS_DIR . '/all.scss'));
-    file_put_contents(self::CSS_DIR . '/all.css', $css);
+    $all_css = $scssCompiler->compile(file_get_contents(self::SCSS_DIR . '/all.scss'));
+    file_put_contents(self::CSS_DIR . '/all.css', $all_css);
+
+    $yt2_css = $scssCompiler->compile(file_get_contents(self::SCSS_DIR . '/yt2.scss'));
```
kauffj: Please call this just "youtube" instead of "yt2". It was initially called yt2 to prevent a conflict with the existing youtube page. Any references to yt2 should just be "youtube" at this point.

view/template/acquisition/youtube.php
```diff
@@ -44,7 +44,7 @@
 
     <div class="content">
       <?php
-      if (isset($_GET['error'])): echo "<div>" . "The following error occurred: ". $_GET['error_message']  . " For support please send an email to hello@lbry.io" . "</div>";
+      if (isset($_GET['error']) && $_GET['error_message']): echo "<div>" . "The following error occurred: ". $_GET['error_message']  . " For support please send an email to hello@lbry.io" . "</div>";
```
kauffj: This check should also happen in the controller similar to the way I highlighted above:
```

  public static function executeXXX()
  {
    return ['acquisition/youtube', [
      'errorMessage' => isset($_GET['error']) && $_GET['error_message']) ? $_GET['error_message'] : null
    ]];
  }

Comment (kauffj) [reply view/template/acquisition/youtube.php]: Above code could be improved via coalesce operator `??`

File: web/js/yt2/youtube_video.js [hunk 4]
```diff
@@ -1,31 +1,25 @@
 
 function playVideo1(){
```
kauffj: This function should not be repeated 3 times.

(Should have caught this one earlier)

Issue #403, opened 2018-02-28
Add redirects for failed google crawls


kauffj:
[lbry-io_20180228T162716Z_CrawlErrors.csv.zip](https://github.com/lbryio/lbry.io/files/1768010/lbry-io_20180228T162716Z_CrawlErrors.csv.zip)

kauffj:
@tzarebczan redirects are currently enumerated here:

https://github.com/lbryio/lbry.io/blob/fda77f620414a59b8058e1dbb444f73a74920bed/controller/Controller.class.php#L132

I suspect a community member could handle reviewing the failed pages and determining where they should go.

kauffj:
Related: https://github.com/lbryio/lbry.io/issues/404

PR #462, opened 2018-03-22
Create .htaccess for 301 Redirects

Smurphonomics:
.htaccess redirects are the prefered and fastest method for smaller amounts of redirects <1k.

All 404's are listed in the file below.

I've only added the redirects that need redirecting, to the .htaccess file. The others require fixing in-page and the case-sensitive broken URL's should be fixed by making permalinks not case- sensitive, rather than using redirects.

[404 Redirects with Notes](https://docs.google.com/spreadsheets/d/1eoTUBlGG_Icn9c22grpeGH5Fx--5REp-tmMRAtWiRHk/edit?usp=sharing)

*This will need checking as I did it at 3:30am while falling asleep.

Related issue: [#403](https://github.com/lbryio/lbry.io/issues/403)


nikooo777:
the website is running on nginx therefore .htaccess are not used/supported,
i'm considering apache2 for a future upgrade, so we can still consider merging this if it makes sense.

tzarebczan:
@nikooo777 is there a way to do something similar in nginx? We could create manual redirects for some of the more common ones in https://github.com/lbryio/lbry.io/blob/master/controller/Controller.class.php#L1366

Smurphonomics:
I can convert to nginx when back at the workstation.

nginx redirects are a lot quicker. :)

Smurphonomics:
@tzarebczan 

Temp redirects in the [code you posted above](https://github.com/lbryio/lbry.io/blob/master/controller/Controller.class.php#L1366) are `307`. This needs a review.

` foreach ([307 => $tempRedirects, 301 => $permanentRedirects] as $code => $redirects)`

`307` redirects are for emergency redirects, such as when a server goes down etc.

Temp redirects should be `302` or `303` when receiving POST data from a client.

kauffj:
I agree that the redirect logic is poor and ought to change.

However, I'd like redirects to be remain in source code whatever the solution is.

I understand that redirects at the server level are more performant, but this is a performance concern that doesn't really matter.

Smurphonomics:
I will convert **redirects** to work in current **source code** ASAP.

I'll also change **307** to **302** unless there was a specific reason for using a **307**?

[What's the difference between a 302 and a 307 redirect?](https://www.pmg.com/blog/301-302-303-307-many-redirects/)

kauffj:
Thanks @Smurphonomics! When I wrote in "source code" I actually just meant in source control. I don't care if the redirects are specifically handled in PHP or not.

Re: 307 vs 302 I think the intention was that the URLs listed as 307s should continue to be used. If that was not correct, you can change to 302 (or 303?).

Smurphonomics:
I'll add the redirects to the existing PH, this might be easier for when and if you change to Apache or use the config file for redirects in the future. Having multiple redirects setups might lead to conflicts and confusion latter on down the line.

I think it's best for 307 to be 302 as 307 is literally for like 1 hour in an emergency.

Smurphonomics:
New PR for redirect is here: #495

PR #480, opened 2018-03-28
Add youtube status analytic

maximest-pierre:


nikooo777:
This will likely trigger the analytics board more than once if the user doesn't have the cookie but comes back through the URL received via email or previously saved. Should we worry about this and try to find a better solution or is it negligible? @finer9

kauffj:
@nikooo777 discussed in Slack. I agree with (and recognized) this downside, but we're okay with it for now.

PR #482, opened 2018-03-28
add redirect for /youtube/status/

### LINKED ISSUE #479: 404 on youtube/status when there is no token
when visiting the /youtube/status page the user gets a 404 not found. The correct behaviour will be if the user has a youtube_token in the local storage redirect the user to the youtube/status page if not redirect to /youtube.

maximest-pierre:
fix #479 

If the user type /youtube/status and has a valid status_token redirect to the token status page. if not then return to /youtube.


nikooo777:
LGTM

PR #492, opened 2018-03-31
Reworded + Corrections - 159-lbry-development-community-update-Jan-20…

Smurphonomics:
…18.md

**Marketing Psychology:** A rewards program is a job or action a person participates in, in order to receive a reward.

When the participant fulfils a task, the reward is **not free**. It is compensation for their work they have carried out.

Calling the reward **free**, after someone has given their time and effort to engage in a promotion, can be seen as insulting and leave a negative impression on the brand offering the reward.

However, when advertising a rewards program, many companies will say participants are rewarded with something for **free**. Technically, this is incorrect for the above reason, but in this case, it leaves a positive impression and acts as an effective carrot on the stick.


tzarebczan:
Thanks @Smurphonomics ! Had to revert 2 lines - you added "santatoshi" randomly and removed it from another place :)

PR #495, opened 2018-03-31
New PHP Redirects

Smurphonomics:
After redirects are checked, published and the cache is cleared, they need to be marked as fixed in GSC (Google Search Console).

Many redirects in the original list have not been included as they look like on page errors.

If you can provide me with the 404 source list (available in GSC) I can check and fix these pages.

Other redirects not included are the case-sensitive ones. Permalinks shouldn't be case-sensitive. These 404 pages will be fixed automatically when the permalinks are no longer case sensitive.

### DIFF
--- a/controller/Controller.class.php
+++ b/controller/Controller.class.php
@@ -153,6 +153,53 @@ class Controller
       '/faq/Q1-17-CreditReport'             => '/credit-reports/2017-Q1',
       '/faq/how-to-report-bugs'             => '/faq/support',
       '/faq/make-money'                     => '/faq/earn-income',
+
+//     New Redirects
+      
+      '/news/$1.2b-market-cap-we-dont-care' => '/news/1.2b-market-cap-we-dont-care',      
+      '/news/lbry-promo-video-raw-footage'  => '/news',																		
+      '/developer-program'                  => '/news',																	
+      '/news/page/2'                        => '/news',																																	
+      '/youtube/sync'                       => '/faq/youtube',																		
+      '/nothing-here'                       => '/news',														
+      '/news/www.maidsafe.com'              => '/faq/different-storj',															
+      '/news/www.porcfest.com'              => '/news/lbry-at-porcfest-first-public-screening-film-via-blockchain',																		
+      '/news/author/jeremy/page/2'          => '/team',																	
+      '/news/every-major-tech-company-hates-you' => '/news',																		
+      '/navframe?selectedItem=/news'        => '/news',																																	
+      '/news/author/jack-robison'           => '/team',																	
+      '/news/mysterious-'                   => '/news/mysterious-100k-lbc-revealed',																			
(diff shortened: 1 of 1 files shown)


controller/Controller.class.php
```diff
@@ -153,6 +153,53 @@ protected static function getRouterWithRoutes(): \Routing\RouteCollector
       '/faq/Q1-17-CreditReport'             => '/credit-reports/2017-Q1',
       '/faq/how-to-report-bugs'             => '/faq/support',
       '/faq/make-money'                     => '/faq/earn-income',
+
+//     New Redirects
+      
+      '/news/$1.2b-market-cap-we-dont-care' => '/news/1.2b-market-cap-we-dont-care',
```
kauffj: What if we move all of this to the 404 route / handler so that they don't have to be declared as routes?

kauffj:
@jsigwart it sounds like if we share our top 404s with @Smurphonomics he'd be willing to update the site with redirects to where they would go. Can you share it? I'm fine with it being posted on this ticket or another one, doesn't need to be secret.

tzarebczan:
@kauffj I believe that's what Stephen already did based on https://github.com/lbryio/lbry.io/issues/403 ( see his attachment in the original post) .  

He was still asking for: If you can provide me with the 404 source list (available in GSC) I can check and fix these pages.

Smurphonomics:
Yep, just to clarify, GSC lists where the 404 was linked from. Many of these are internal links where the link has errors. Sometimes, this is a plugin, sometimes this is a typo.

I can use this list to locate the pages to fix.

controller/Controller.class.php
tzarebczan: @kauffj who can work on this? @maximest-pierre ?

tzarebczan:
@Smurphonomics has access to the data now, he'll work on identifying the source pages.

Otherwise, @kauffj had a comment in the review section to possibly move the routes to a 404 route handler. How do we proceed?

kauffj:
@tzarebczan yes, @maximest-pierre can finish this PR. I'd like two more changes:

- Don't use the routing class to perform the redirects. Instead perform the redirects after routes don't match. Either inside of the 404 action or as a new separate method that happens in between routing and sending the user to the 404 action on routing failure.
- The redirects to be cleanly separated in a way that is user-friendly for non-technical people to edit them. Something like `data/redirects.txt` would probably work fine.

maximest-pierre:
I am going to get to this PR later in the week

tzarebczan:
@maximest-pierre Just a friendly bump. If you have higher priority items, focus on those :)

kauffj:
@maximest-pierre I added this to this week's sprint since I hate lingering PRs and this is so close to deployable. Please tackle this after email SQL triggers or if you get stuck on SQL triggers and want a quick hit.

maximest-pierre:
#552

PR #496, opened 2018-03-31
fix error on contact us page

ykris45:
screenshot_2018-03-31-16-59-23-657
i saw error when access lvry.io/learn  maybe this can be fix


tzarebczan:
Thanks @ykris45, @jsigwart broke it... Tsk tsk lol

PR #494, opened 2018-03-31
Custom SCSS File Import

Smurphonomics:
Imports the new custom SCSS file.

Required for changes to the [News ](https://lbry.io/news) page. See #493

PR #497, opened 2018-03-31
News ID

Smurphonomics:
Always good to add a relatively named **ID** for each page so it can be targeted.

*This **ID** is required for the **CSS** changes I made to the news page. #493

### DIFF
--- a/view/template/content/news.php
+++ b/view/template/content/news.php
@@ -1,6 +1,6 @@
 <?php Response::setMetaDescription(__('description.news')) ?>
 <?php echo View::render('nav/_header', ['isDark' => false]) ?>
-<main>
+<main id="news">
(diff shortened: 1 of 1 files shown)


maximest-pierre:
PR #493 #494 and #497  should be combined into a single PR. Also, the custom.scss should be removed and the content moves to _content.scss. You can also define colours in global.scss. Once everything is inside a single PR I am going to look into this further.

Smurphonomics:
Still learning GitHub - So not sure how to make everything 1 pull request atm.

The reason for the custom.scss file was to have an easy-to-find place for targeting specific areas without affecting others while keeping code to a minimum.

I will review the _content.scss and see if putting code in there would be more optimal.

tzarebczan:
@Smurphonomics closing this, let me know if you still need help with github/making changes in a single branch.

PR #520, opened 2018-04-26
Updated total amount credits in summary

DiSae:
Updated 'We spent 393,525 total community credits on line items detailed below.' to 'We spent 393,585 total community credits on line items detailed below.' as the sum of the items below the summary is 393,585.


tzarebczan:
@finer9 can you confirm and merge?

PR #524, opened 2018-04-29
Add pre commit hook for automatic code fix

maximest-pierre:
This PR adds a pre-commit hooks that automate the linting process.


kauffj:
@maximest-pierre when this is ready, please separate the commit to simply add the hook from the commit that runs the hook and introduces a massive amount of formatting changes.

maximest-pierre:
@kauffj Its ready to be look at

kauffj:
Would like @lyoshenka to review this one instead of me.

dev.sh
```diff
@@ -6,13 +6,16 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
+ cp "$DIR/data/config.php.example" "$DIR/data/config.php"
 fi
 
 if ! which $PHPBIN 2>/dev/null; then
-    PHPBIN=php
+   PHPBIN=php
 fi
 
+# Installing git hook
+./hooks/install.sh
```
lyoshenka: please use $DIR instead of `.`, since the current directory may not be what you think

docker.sh
```diff
@@ -4,14 +4,18 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 
 if [ ! -e "$DIR/data/config.php" ]; then
-  cp "$DIR/data/config.php.example" "$DIR/data/config.php"
+ cp "$DIR/data/config.php.example" "$DIR/data/config.php"
 fi
 
+# Installing git hook
+./hooks/install.sh
```
lyoshenka: same here

maximest-pierre:
Testing the pre-commit hook for the last two days and everything seems fine.

PR #525, opened 2018-04-29
Add cors to dmca page

maximest-pierre:
screenshot from 2018-04-29 12-10-54

### DIFF
--- a/controller/action/ReportActions.class.php
+++ b/controller/action/ReportActions.class.php
@@ -4,6 +4,7 @@ class ReportActions extends Actions
 {
   public static function executeDmca()
   {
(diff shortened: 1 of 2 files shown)


view/template/report/dmca.php
```diff
@@ -1,3 +1,4 @@
+<?php header("Access-Control-Allow-Origin: *"); ?>
```
kauffj: I'm going to be opaque in my response for teaching a man to fish purposes.

How are headers set elsewhere in the site? If you don't know, how would you go about answering that question?

maximest-pierre: Response::setHeader("Acces-Control-Allow-Origin", "*");

PR #536, opened 2018-05-03
Change roadmap item display

### LINKED ISSUE #530: Roadmap enhancements/fixes
Discussed on Growth call today - https://lbry.io/roadmap

1) Items from Q1 still show as future, we may want to show these after the future items

2) the "other" tag looks like it's coming from the badge, which is the project name: https://github.com/lbryio/lbry.io/blob/b36b3fbc27b0c47883880794add97ce25e9c8f5d/lib/thirdparty/Asana.class.php#L46
-@kauffj - did we still want to assign projects in Asana to these (my guess is no?), or can we re-purpose this as a Status somehow in Asana? Status would have Not Completed, Partially Completed, Completed.

Did I miss anything?

maximest-pierre:
fixes #530

### DIFF
--- a/lib/thirdparty/Asana.class.php
+++ b/lib/thirdparty/Asana.class.php
@@ -14,6 +14,8 @@ class Asana
       161514803479899 => ['Blockchain and Wallets', 'https://github.com/lbryio/lbrycrd'],
       136290697597644 => ['Integration and Building', null],
       158602294500249 => ['Documentation', null],
+      658477315612493 => ['Complete', null],
+      658477315612495 => ['In progress', null]
     ];
 
     $tasks = [];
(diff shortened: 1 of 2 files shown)


lib/thirdparty/Asana.class.php
```diff
@@ -14,14 +14,17 @@ public static function listRoadmapTasks($cache = true)
...
+      658477315612495 => ['In progress', null]
     ];
 
     $tasks = [];
 
     $tags = [
       192699565737944 => 'Open Beta',
       542803886522122 => 'Upcoming',
-      542803886522120 => 'Future'
+      658477315612491 => ' 2018 ',
```
maximest-pierre: Had to leave a space to prevent php from converting to int and changing it to zero. Even using (string) has not fixing it.

PR #537, opened 2018-05-04
update team photo to newest

ykris45:
i download this photo via LBRY Twitter and would be nice if lbry.io/join-us always updated


tzarebczan:
@ykris45 I'll have someone on the team find a funnier photo to use...it's supposed to be quirky :p

PR #543, opened 2018-05-08
Formatted Page With Extras

Invariant-Change:
Just gave the page a cleanup added "**OPEN IN LBRY**" buttons (with open links) and made the markup compatible with **accessibility devices**.

Although _multiple_ **H1 tags** can be used, they are **not** _best practice_ due to accessibility **issues**.
image

### DIFF
--- a/view/template/page/3d-printing.php
+++ b/view/template/page/3d-printing.php
@@ -2,50 +2,38 @@
 <?php Response::setMetaTitle(__('title.meetup')) ?>
 <?php echo View::render('nav/_header', ['isDark' => true, 'isAbsolute' => true]) ?>
 
-<main>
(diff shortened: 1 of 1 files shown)


tzarebczan:
@jamesbiller can you run this locally to see if it's something you'd like merged.

jamesbiller:
You got it chief!

Invariant-Change:
Currently working blind as I don't have lbry.io set up on a local server, but all should hopefully look like I imagined :)

Invariant-Change:
Oh and forgot to say... Images sizes should always be in HTML. This will help speed up page load times and make a page compatible with every device/browser.

Invariant-Change:
Still learning Git, so looking to see what happened.

PR #544, opened 2018-05-09
Update 3d-printing.php

Invariant-Change:
Fixed: Images sizes and file type.

Images for the site should be much much smaller and jpg for these types of images unless transparency is needed.

**PNG**: Block graphics such as logos.
**JPG**: Gradient graphics such as photos.

New images are jpeg compressed and resized to 500x500, sRGB, progressive.

Typical size for an image of **500px by 500px** would be something like **10kb-35kb.** previous images were in excess of 2.3mb each. New image sizes are around 20kb.

Large images would not work well on mobile devices and would have a heavy **impact** on **search engine ranking**. Also, browser resizing can make lage images look pixelated or blurry.

image.

Gave images a little clean up too - OCD 😄 

image

- Alt tags corrected. Had missing `""` 
- Title tags added.
- Corrected LBRY Alt tag which was set as Dash.
- Correctly nested HTags
- Used Sections where appropriate.
- Corrected spelling grammer mistakes.
- Converted CSS width/height to HTML.

*If `H1, H2, H3` styling isn't the look you are going for, use CSS to style rather than using incorrectly nested H Tags.

I typically use Bootstrap for standard base styles, then use a separate custom CSS file on top to be more flexible in various areas where needed.

*Still working blind, but should be OK.

### DIFF
--- a/view/template/page/3d-printing.php
+++ b/view/template/page/3d-printing.php
@@ -1,40 +1,31 @@
+
 <?php Response::setMetaDescription(__('description.3d-printing')) ?>
 <?php Response::setMetaTitle(__('title.3d-printing')) ?>
 <?php echo View::render('nav/_header', ['isDark' => true, 'isAbsolute' => true]) ?>
 
-<main>
-  <div class="cover cover-dark cover-center cover-full cover--dark-overlay" style="background-image: url(/img/3d-crypto-accepted-here.jpg)">
+<main style="text-align:center;">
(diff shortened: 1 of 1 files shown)


tzarebczan:
@jamesbiller this is the updated PR from Stephen :)

jamesbiller:
@Invariant-Change Accepted some of your changes outside of this as the code didn't work completely, but I did accept your changes to the images and the text boldens.

PR #546, opened 2018-05-09
A Tidy Up

Invariant-Change:
Took out periods from lists (not required and looks better without them).
Took out commas from the "Join us on" section. They didn't look right.

image

image


tzarebczan:
@Invariant-Change if we are removing periods, we should be consistent across the rest of the site. Also wouldn't we remove it from "Experience digital abundance" ?

Invariant-Change:
"Experience digital abundance" is brought in dynamically from another place, not editable on that page.

There is only one footer, so that would be consistent. Typically, footer lists/menus don't use periods.

http://www.businesswritingblog.com/business_writing/2012/01/punctuating-bullet-points-.html

PR #552, opened 2018-05-11
File based redirect

maximest-pierre:
Add yaml based redirect instead of hardcoding array.


kauffj:
@maximest-pierre this looks good, but can we move the redirection checking so that it happens after routing fails?

That is, rather than add a route for each redirection, only do the redirection logic when no match is found.

This will be a more performant and scalable approach.

PR #553, opened 2018-05-11
add lbry is hiring on footer

ykris45:
Results

screenshot_2018-05-12-04-30-32-741
based on Mr Jeremy last comment . in my closed PR


tzarebczan:
@ykris45 Jeremy already added that to the code :)

PR #555, opened 2018-05-14
fix minor for reddit tipbot FAQ

ykris45:


tzarebczan:
Good catch, thanks @ykris45 !

Issue #561, opened 2018-05-16
Add link to shop.lbry.io on the website


maximest-pierre:
One of the perfect place for it would be in the footer.

PR #562, opened 2018-05-16
add lbry shop on footer

ykris45:
based on this issues #561

### DIFF
--- a/view/template/nav/_footer.php
+++ b/view/template/nav/_footer.php
@@ -19,6 +19,9 @@
       <div>
         <a href="/join-us">LBRY is hiring!</a>
       </div>
+      <div>
+        <a href="https://shop.lbry.io">LBRY Shop</a>
+      </div>
     </div>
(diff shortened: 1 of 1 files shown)


tzarebczan:
@ykris45 can you add a screenshot of how this looks?

ykris45:
screenshot_2018-05-17-21-26-15-202
@

tzarebczan:
@ykris45 that's definitely not what we want :) 

@kauffj  / @maximest-pierre any thoughts on where/how this should live in the footer?

kauffj:
I'm not sure it should :)

Though /learn seems in appropriate as well.

I suppose I don't object to it being after "News" in the footer and before the icons.

tzarebczan:
@ykris45 closing this for now, see comments above if you want to give this another shot. Sounds like adding it to the learn page and footer after news. Maybe you can find an icon to go with it.

PR #566, opened 2018-05-18
Update script

maximest-pierre:
- Update dev.sh to stop on error
- Add an additional note for php extension

### DIFF
--- a/dev.sh
+++ b/dev.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 PHPBIN=php7.2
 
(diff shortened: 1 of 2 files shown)


dev.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
lyoshenka: i like this. if you want to go full bash strict mode, see http://redsymbol.net/articles/unofficial-bash-strict-mode/

lyoshenka:
LGTM, forgot to click the approve  button

PR #569, opened 2018-05-21
Youtube versions

maximest-pierre:


controller/action/AcquisitionActions.class.php
```diff
@@ -27,14 +27,24 @@ public static function executeYouTubeSub()
...
             $error_message = Request::encodeStringFromUser($_GET['error_message']);
         }
 
-        return ['acquisition/youtube', [
+        $baseTemplate = 'acquisition/youtube';
+        $versionedTemplate = $baseTemplate . '-' . $version;
+        $template = $version && View::exists($versionedTemplate) ? $versionedTemplate : $baseTemplate;
+
+        if ($version && View::exists($versionedTemplate)) {
+            Session::set("template", $template);
```
kauffj: "template" needs to be a const and probably shouldn't be named so simply (session values are global).

controller/action/AcquisitionActions.class.php
```diff
@@ -27,14 +27,24 @@ public static function executeYouTubeSub()
...
+        }
+
+        $template = Session::get("template") ?? $template;
```
kauffj: What if I have a template in my Session that no longer exists or got moved?

PR #573, opened 2018-05-24
Update INSTALL.md

ykris45:
because the repository needs composer

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,6 +7,7 @@ To run this project, you'll need to have either PHP7 or Docker installed, and be
 ## Running Locally
 
 - Install PHP7
+- Install [Composer](https://getcomposer.org/download/) 
 - Checkout the project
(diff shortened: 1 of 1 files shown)


tzarebczan:
@maximest-pierre can you confirm composer needs to be installed? I haven't tried stepping through since the last time it was failing.

maximest-pierre:
No composer does not need to be install on the system. If you look the .phar package is already at the root and all the script are using that one and not the local version. I tested it a week ago when we discovered that the ./dev.sh script was acting weird.

tzarebczan:
@ykris45 can you try this on a clean VM for Windows/Linux somewhere? Did you ever get a chance to setup virtuabox? Might be tough with your low memory :\

ykris45:
@tzarebczan i doesnt have Virtual Machine .. so its based on your DM earlier

PR #574, opened 2018-05-24
Update composer to version 1.6.5 from 1.5.2

maximest-pierre:
Update composer binary to the latest version


tzarebczan:
@maximest-pierre was this tested locally? Does it need to be reviewed?

ykris45:
@tzarebczan this is just update binary for composer to the newest version. and same with jquery before.

maximest-pierre:
It was tested locally. To test it you only need to clone the repo in a temp directory and run the ```dev.sh```.

tzarebczan:
LGTM, pushing.

PR #584, opened 2018-05-31
Create 170-forking-for-asics.md

Feanor78:
Trying to figure out how to make the captions below the photos in this blog post show up properly without breaking markdown - after the 2nd <div> markdown stops working, and I'm not sure if the images are showing up right

Haha, I tried to type "<div>" and it created a div


nikooo777:
I think we should keep the images somewhere in our repo here otherwise we're trusting external websites to host them for us. It's safer to just add them here.

Let's jump on a call tomorrow so I can show you

Feanor78:
@kaykurokawa if you merge this PR it'll publish the post to our blog, then you can go ahead and post it on Medium/set the canonical URL to the right location. Thanks!

PR #587, opened 2018-06-02
fork announcement

lyoshenka:


content/news/171-hard-fork-july-2018.md
```diff
@@ -0,0 +1,31 @@
...
+Only claim expiration is changing. The reward schedule, token distribution, hash algorithms, etc. are all staying the same.
+
+The fork will take effect at block 400155, which will occur on July 9th, 2018 at around noon EST.
+
+## What You Should Do
+
+For most people, nothing needs to be done. The fork will take effect transparently and won't change your experience of LBRY.
+
+**If you are running a full node or mining pool, upgrade to the latest version of lbrycrd as soon as possible**.
+Pre-built binaries areavailable on [the releases page](https://github.com/lbryio/lbrycrd/releases). You will need version 0.12.2.0 or higher.
```
lbrynaut: s/areavailable/are available/

content/news/171-hard-fork-july-2018.md
```diff
@@ -0,0 +1,31 @@
+---
+author: alex-grintsvayg
+title: 'Hard Fork on July 9th, 2018'
```
kauffj: I would propose we have some kind of naming convention for hard and soft forks so that we can refer to them canonically.

Right now I would call this something like `hardfork_2018-07-09`, but there may be a better pattern than this. Worth looking at what others do.

view/template/page/forklist.php
```diff
@@ -0,0 +1,25 @@
+<?php Response::setMetaDescription('Subscribe to be notified of fork-related updates') ?>
+<?php Response::setMetaTitle('Blockchain Fork Updates') ?>
+<?php echo View::render('nav/_header') ?>
+<main>
+  <div class="cover cover-dark cover-center cover--dark-overlay" style="background-color:rgb(74, 175, 152)">
```
kauffj: This doesn't have to be a full cover page. I would recommend doing something closer to https://lbry.io/list/subscribe without the sidebar.

lyoshenka: I copied lbry.io/newslist. Should I also change that one?

kauffj: Possibly, but I don't own that page and I'm not sure how it's being used.

content/news/171-hard-fork-july-2018.md
lyoshenka: I have not been able to find any standard for this. I think its unlikely that we'll do more than one fork in a month, so just year and month should be enough. I propose we call this one `HF1807`.

view/template/page/forklist.php
lyoshenka: I'm gonna leave this as is. Someone will redesign these pages eventually anyway.

PR #594, opened 2018-06-04
add FAQ link to DMCA page

ykris45:
to make helpfull to user who has reported infringing content
image


kauffj:
Good one @ykris45 ! Hit up @tzarebczan on Discord for a tip

PR #598, opened 2018-06-06
Update exchanges.md

ceoger:
LBC is also listed on Trade Satoshi
https://tradesatoshi.com/Exchange/?market=LBC_BTC


tzarebczan:
Thanks @ceoger !

PR #599, opened 2018-06-06
Update exchanges.md

ceoger:
i recommend this should be "You can buy or sell LBRY Credits at one of these"  "LBRY Credits" instead of just "credits" if that's not intentional


tzarebczan:
Thanks @ceoger !

PR #606, opened 2018-06-09
Closes #597

### LINKED ISSUE #597: Move the Optimize Page-Hiding Snippet to the head
When I open any lbry.io page, I get the following effect:
- Top of the page loads normally
- Page changes to all white
- Page loads normally again

This happens in a span of 1-2 seconds, creating a flashing effect every time I change the page.

The issue is that the [Optimize Page-Hiding Script](https://developers.google.com/optimize/) you are using is at the bottom of the website, so it takes a while when the browser reaches that part of the page and renders the all white website. You should move this piece of code to the head section so it loads faster.

Krisseck:
Separates the analytics to head and footer parts, to solve issue #597


kauffj:
Perfect, thank you @Krisseck

Issue #504, opened 2018-04-04
case insensitive URLs


tzarebczan:
Thanks to @Smurphonomics for pointing this out: Our URLs should be case insensitive: https://lbry.io/news/lbry-development-community-update-mar-2018 works but https://lbry.io/news/lbry-development-community-update-Mar-2018 does not. 

Might be useful: 
https://stackoverflow.com/questions/18415508/how-to-make-url-case-insensitive-with-nginx

tzarebczan:
Same as https://github.com/lbryio/lbry.io/issues/556 - will close this one as 556 as the conical URL mention.

PR #609, opened 2018-06-10
Case Insensitive Url

### LINKED ISSUE #556: Drop case sensitivity in URLs
lbry.io/reefermadness and lbry.io/ReeferMadness should both show the same page

Additionally, this ticket should not be considered complete without `rel=canonical` meta tag so that both count the same for SEO purposes.

maximest-pierre:
Fixes #556 #504

### DIFF
--- a/controller/Controller.class.php
+++ b/controller/Controller.class.php
@@ -54,7 +54,7 @@ class Controller
             $dispatcher = new Routing\Dispatcher($router->getData());
             return $dispatcher->dispatch($method, $uri);
         } catch (\Routing\HttpRouteNotFoundException $e) {
-            return NavActions::execute404();
+            return static::doLowerUriRedirectOr404($uri, $router);
         } catch (\Routing\HttpMethodNotAllowedException $e) {
             Response::setStatus(405);
             Response::setHeader('Allow', implode(', ', $e->getAllowedMethods()));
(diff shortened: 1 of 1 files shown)


controller/Controller.class.php
```diff
@@ -148,6 +152,10 @@ protected static function getRouterWithRoutes(): \Routing\RouteCollector
                 Response::enableHttpCache();
                 return ['page/' . $slug, []];
             } else {
+                $lower_slug = strtolower($slug);
+                if ($slug !== $lower_slug) {
+                    return static::execute("GET", $lower_slug);
```
kauffj: `Route::GET`

controller/Controller.class.php
```diff
@@ -148,6 +152,10 @@ protected static function getRouterWithRoutes(): \Routing\RouteCollector
                 Response::enableHttpCache();
                 return ['page/' . $slug, []];
             } else {
+                $lower_slug = strtolower($slug);
```
kauffj: DRY violation

controller/Controller.class.php
```diff
@@ -54,6 +54,10 @@ public static function execute($method, $uri)
             $dispatcher = new Routing\Dispatcher($router->getData());
             return $dispatcher->dispatch($method, $uri);
         } catch (\Routing\HttpRouteNotFoundException $e) {
+            $lower_uri = strtolower($uri);
```
kauffj: capitalization doesn't match rest of file

controller/Controller.class.php
```diff
@@ -148,13 +148,22 @@ protected static function getRouterWithRoutes(): \Routing\RouteCollector
...
             } else {
-                return NavActions::execute404();
+                return static::checkLowerUri($slug);
             }
         });
 
         return $router;
     }
 
+    protected static function checkLowerUri($uri)
```
kauffj: This function has unclear side effects. One would not reasonably expect `checkLowerUri` to re-route to a 404.

There is likely a better name than this, but if you can't think of anything better, name it `doLowerUriRedirectOr404`.

kauffj:
@NetOperatorWibby is correct that this still has problems.

Originally I was thinking we'd redirect to the lowercased version, but since we're just re-routing internally, this design results in multiple URLs serving the same content. So either you need to set a flag when this happens and set the meta tag appropriately, or you need to redirect (301) rather than just re-execute routing.

With regards to tokens, this is "not safe" in the sense that theoretically this could cause a spurious match, but it is safe in the sense that multiplying the tokens tried by two should have no real-world consequence given a sufficiently large number/token space (which we should have in all cases).

controller/Controller.class.php
```diff
@@ -148,13 +148,24 @@ protected static function getRouterWithRoutes(): \Routing\RouteCollector
...
         });
 
         return $router;
     }
 
+    protected static function doLowerUriRedirectOr404($uri, $router = null)
+    {
+        $lowerUri = strtolower($uri);
+        if ($uri !== $lowerUri) {
+            if (View::exists('page/' . $lowerUri) || isset($router) && $router.hasroute($lowerUri)) {
```
kauffj: Did you run this? I don't think `$router.hasroute` is valid PHP

maximest-pierre: yes,
In route_collector.class.php which is the base class for $router:
```
public function hasRoute(string $name): bool
    {
        return isset($this->reverse[$name]);
    }
```

kauffj: So you're confidant this code works?

maximest-pierre: yes, I tested 404, dynamic and static routes.

tzarebczan:
Putting this on hold until either @nikooo777 or @NetOperatorWibby can get to it, unless someone else wants to take over.

nikooo777:
@NetOperatorWibby is this issue still valid with the new .io ?

kauffj:
Yes, it's just a new design not new (backend) code.

PR #633, opened 2018-06-15
Update transaction-types.md

ceoger:
Many transaction types also have details associated with them such as the claim/channel name or if they came from (changed "an" to "a") LBRY Reward


tzarebczan:
Thanks Clement!

PR #637, opened 2018-06-16
Text fixes

Grayyyyy:
More text fixes :smile:


tzarebczan:
Thanks again! 

Some of the changes are hard to see using the git diff, not sure if that's how you are changing it, or if that's just normal.

PR #642, opened 2018-06-18
Magic links for all

tzarebczan:
This should effectively revert https://github.com/lbryio/lbry.io/pull/299 as magic links should now work on Windows.


tzarebczan:
@maximest-pierre you should be able to change the user agent in chrome. In terms of the link itself, I've tested that on Windows.

PR #647, opened 2018-06-25
Austen

tzarebczan:
Might need a title change
Needs to be proofread/formatted 
Julie to work on Cover
major/minor version discussion?


tzarebczan:
@Feanor78 / @jamesbiller one of the top items should now be: 
Significantly faster app experience for new and existing users with upgraded daemon - https://github.com/lbryio/lbry/releases/tag/v0.20.2 - we could mention that new user home page load went from 1 minute to 5 seconds. And channel queries went from 30 seconds to 3.

tzarebczan:
Once we have a download link, will also need to be included.

PR #656, opened 2018-06-26
Update shapeshift.md

ceoger:
Updated the pics with the new app design, also edited steps to convert and verify that LBC is received


tzarebczan:
@ceoger please fix the spelling error in the last screenshot per my Discord message.

PR #659, opened 2018-06-27
set screenshot of lbry io homepage to newest

ykris45:


ykris45:
@jsigwart published is disabled on speech.

Issue #527, opened 2018-05-01
Make sure repo/install is up to new documentation guidelines


tzarebczan:

tzarebczan:
Still need to run through install procedure to make sure we have all the steps documented.

tzarebczan:
When trying to run the site on a clean Ubuntu 17.04 machine after running `sudo apt-get install php`, I see the following: 
```
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - The requested PHP extension ext-curl * is missing from your system. Install or enable PHP's curl extension.
  Problem 2
    - The requested PHP extension ext-mbstring * is missing from your system. Install or enable PHP's mbstring extension.
  Problem 3
    - The requested PHP extension ext-xml * is missing from your system. Install or enable PHP's xml extension.
  Problem 4
    - Installation request for erusev/parsedown 1.7.1 -> satisfiable by erusev/parsedown[1.7.1].
    - erusev/parsedown 1.7.1 requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
  Problem 5
    - Installation request for phar-io/manifest 1.0.1 -> satisfiable by phar-io/manifest[1.0.1].
    - phar-io/manifest 1.0.1 requires ext-dom * -> the requested PHP extension dom is missing from your system.
  Problem 6
    - Installation request for phpunit/php-code-coverage 5.3.0 -> satisfiable by phpunit/php-code-coverage[5.3.0].
    - phpunit/php-code-coverage 5.3.0 requires ext-dom * -> the requested PHP extension dom is missing from your system.
  Problem 7
    - Installation request for phpunit/phpunit 6.5.7 -> satisfiable by phpunit/phpunit[6.5.7].
    - phpunit/phpunit 6.5.7 requires ext-dom * -> the requested PHP extension dom is missing from your system.
  Problem 8
    - Installation request for theseer/tokenizer 1.1.0 -> satisfiable by theseer/tokenizer[1.1.0].
    - theseer/tokenizer 1.1.0 requires ext-dom * -> the requested PHP extension dom is missing from your system.
  Problem 9
    - erusev/parsedown 1.7.1 requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
    - erusev/parsedown-extra 0.7.1 requires erusev/parsedown ~1.4 -> satisfiable by erusev/parsedown[1.7.1].
[...]

PR #663, opened 2018-07-05
add php download links

ykris45:
probably should link to https://www.docker.com/get-docker for install docker section


tzarebczan:
Let me know how you make out with the install, you may need to do this : https://github.com/lbryio/lbry.io/issues/527#issuecomment-393647193

It will vary by distro or if you are on Windows. Haven't had a chance to update the docs.

ykris45:
@tzarebczan its from Php official site . http://php.net/manual/en/install.php . im install php on windows machine using the method from that site and success

ykris45:
https://www.sitepoint.com/how-to-install-php-on-windows/

PR #665, opened 2018-07-06
Update tipbot-twitter.md

ceoger:
removed the magic word `lbryian` and made some changes to the tweet links, withdraw seems to require the amount before it works, so i updated that as well. withdraw command alone does not work example; https://twitter.com/TrendsPremium/status/1015249162445697025


tzarebczan:
Thanks @ceoger !

PR #678, opened 2018-07-12
embed official lbry tweet to this news blogpost

ykris45:


tzarebczan:
nice addition @ykris45 , I moved it a bit higher in the article.

PR #681, opened 2018-07-19
Update exchanges.md

ceoger:
added https://www.simpleswap.io/ to the exchanges where LBC can be bought


tzarebczan:
Thanks @ceoger ! Modified it to support the new format (traditional vs instant)

PR #686, opened 2018-07-25
Update _youtube_header.php

ykris45:
add backlink to logo


ykris45:
@tzarebczan i want to close this.. because out of memory

PR #687, opened 2018-07-25
delete closed PR on ledger status

ykris45:
To make this always updated
image
and add dot to make looks preety


tzarebczan:
Still want to keep this on the page since it will need to be revisited.

ykris45:
@tzarebczan how to add - as dots to make this page more pretty and clearly

PR #693, opened 2018-07-30
add screenshot of LBRY rewards

ykris45:


tzarebczan:
We should have some new rewards + descriptions tomorrow, let's update this at that time. Will need to update the reward list also.

tzarebczan:
Thanks Chris, replaced the screenshot and updated the descriptions

PR #701, opened 2018-08-07
Update youtube.md

ceoger:
made it clear that, users won't see their channel/content in their LBRY app under `My LBRY - Publishes` until LBRY sends their wallets/claims over


tzarebczan:
Thanks @ceoger !

PR #707, opened 2018-08-13
Update samuel-lbryian.md

ykris45:
added twitter and github backlink to official lbry.io pages


ykris45:
@tzarebczan Thanks Sir

PR #710, opened 2018-08-13
 Add Yt sync page

ykris45:
image

my suggestion to gain more youtubers sync with LBRY Protocol


tzarebczan:
@robvsmith / @kauffj what do you think about this addition?

kauffj:
@tzarebczan it ought to be added to this page but not precisely in this fashion. I'll review and fix.

ykris45:
image
i had other ideas but ya just share my idea...hehe

kauffj:
Addressed in https://github.com/lbryio/lbry.io/commit/2697cd65b95c3ebf01ad279f11c0bb9bec777c38 but I still dislike this. Opening separate private ticket related to this.

PR #711, opened 2018-08-14
temp changes

tiger5226:


lib/thirdparty/LBRY.class.php
```diff
@@ -25,6 +25,21 @@ public static function subscribe($email, $tag = null)
     ]), ['json_response' => true]);
     }
 
+    public static function editEmailSettings($token, $email, $isPrimary =null, $isEnabled = null)
+    {
+        return Curl::post(static::getApiUrl('/user/email/edit'),['auth_token' => $token],['email' => $email],['is_primary' => $isPrimary],['is_enabled' => $isEnabled]);
```
kauffj: Shouldn't the last 3 parameters to this call be wrapped in an array? Same for the calls below this, too.

view/template/mail/_settingsForm.php
```diff
@@ -0,0 +1,12 @@
+<?php $error = $error ?? null ?>
+<?php $tag = $tag ?? null ?>
+<?php $largeInput = $largeInput ?? false ?>
+<form id="settings_form" action="/list/subscribe" method="POST" novalidate>
+  <?php if ($error): ?>
+    <div class="notice notice-error spacer1"><?php echo $error ?></div>
+  <?php endif ?>
+  <div class="mail-submit" >
+    <input name="receive" type="radio" value="true" checked><label>Receive Email</label><br>
```
kauffj: Should probably just be a single submit button, though I suppose there could be a cancel button too.

tiger5226:
@kauffj - It works end to end now. Also the email management api is merged into internal-apis. So you can now use this branch with internal-apis without running a local instance.

Running locally:

`http://localhost:8000/list/edit/<auth_token>`

I also added some styling because I just could not take those default checkboxes. I hope you like it, I used the LBRY green. 

screen shot 2018-08-18 at 11 45 18 pm

I think the code probably needs to be cleaned up. I am not great with javascript, jquery but I will try to clean it up some more once I get feedback from you.

Issue #531, opened 2018-05-02
lbry.io mobile version issue


ceoger:
There is an issue on the mobile version of the website, at the very lower button, below the what's next, the Get LBRY button is separated, making the LBRY barely visible. 
Url https://lbry.io/news/lbry-development-community-update-mar-2018

tzarebczan:
@ceoger thanks for filing this, sent you a few LBC on Discord :)

ykris45:
#715

ykris45:
@tzarebczan close close... hehehe.

PR #715, opened 2018-08-15
alternative to fix mobile issue

ykris45:
set height it will looks like tis screenshot
image

### DIFF
--- a/web/scss/_basic.scss
+++ b/web/scss/_basic.scss
@@ -170,6 +170,7 @@ a:hover img
 .btn-full-width
 {
   width: 100%;
+  height: 100%;
 }
 .no-label .btn-label
 {


ykris45:
screenshot_2018-08-16-03-28-38-989
Mobile View @tzarebczan

ykris45:
Desktop Preview
image. if use mobile version its using height on the css and if using desktop version its using width on the css class. arternative method

tzarebczan:
This is an attempt to fix https://github.com/lbryio/lbry.io/issues/531 

I'll let @NetOperatorWibby take a look, maybe he has other ideas.

ykris45:
yessir.. 
image
For Now the Css class for this button is only applied to the _learnfooter.php and it can't broke any css element on the lbry.io site @tzarebczan ..

ykris45:
@NetOperatorWibby Thanks Mr Paul.. Gbu

PR #728, opened 2018-08-25
Update 3d-printing.php

ykris45:
added Youtube Embed video , Fix minor typos, Add Class link Primary and change staging link to spee.ch instead because staging is an test server

image


tzarebczan:
@jamesbiller can you review these changes?

PR #729, opened 2018-08-25
change staging link to spee.ch instead

ykris45:
staging is a speech test server, and i think it will shut down in the future, because test server. and the best use is spee.ch ( live version) and never turned off in the future..


tzarebczan:
Thanks Chris!

PR #758, opened 2018-09-13
Fix multi email submits

tiger5226:
@kauffj - I actually work with ShawnK/Sean to resolve this issue. We use the `Promise.all` to be able to execute multiple api calls and check if one of them failed. It works great from my testing.


tiger5226:
Jeremy gave verbal approval for merger at all hands.

PR #759, opened 2018-09-13
Added descriptions per Rob for reward, subscription tag.

tiger5226:


tiger5226:
jeremy gave verbal approval for merger

PR #760, opened 2018-09-14
Update for small format: wallet -> `Wallet`

namujan78:
Original/Before:
- Click on the wallet tab on the LBRY App `1`

Changed/After:
- Click on the `Wallet` tab on the LBRY App `1`


tzarebczan:
Thanks @namujan78 !

PR #769, opened 2018-09-17
fix link download

ykris45:


tzarebczan:
@ykris45 the download link is correct, it goes to a beta testing opt in page so users are aware this is very beta software.

PR #775, opened 2018-09-17
Update how-to-change-email.md

ceoger:
added android version of changing email


tzarebczan:
Merging for now, but these steps will change once we let users manage their emails via lbry.io (coming soon).

PR #786, opened 2018-09-21
update supply info

tzarebczan:
Add note to clarify why the value was updated.


kauffj:
If you want to provide a historical record of this update, just use Git commit history.

PR #804, opened 2018-10-04
added lbry.tech

ykris45:
image


tzarebczan:
I'll be revamping the developer area where lbry.tech should go, no need to do anything here for now Chris, thanks!

PR #813, opened 2018-10-09
Delete press-kit.md

robvsmith:
Deleting the press kit page, its very out of date, inaccurate, and if there is a refresh for this page coming it will need the branding document finished.

Is there anything this is going to break that jumps out at you @kauffj ?


kauffj:
There's additional logic that needs to be deleted, as-is this will result in the page erroring.

I'll delete the whole thing, though.

robvsmith:
Sounds good, thanks. Is there a place in KB I can read up on the additional logic so I know better how these pages interact with the site?

kauffj:
@robvsmith not really. The press-kit logic is pretty custom, it takes all of the team bios and stuffs them into a zip, along with the images located in `img/press`.

The logic is [here](https://github.com/lbryio/lbry.io/blob/05defac6d1780049606a66174b3ad21a9c13be21/controller/action/ContentActions.class.php#L253) but I wouldn't expect you to understand or find that.

jsigwart:
I'd like to see this information updated on lbry.io. Not sure if all of it needs to live on this particular page (I'm not against deleting the page and moving info elsewhere), as the brand kit will be downloaded from the new brand guidelines & assets page. It is a good thing to have big news articles in a place where folks can see them as it gives credibility to the project.

kauffj:
The press-kit page has been replaced with a prompt to contact Brinck: https://lbry.io/press-kit

I think @Feanor78 has been maintaining a sheet with press mentions, but let's not have this conversation on this PR.

A press-kit and sharing this information publicly has a lot of merit, but having a page up as outdated as this one was was clearly wrong.

PR #814, opened 2018-10-09
embed hackathon

ykris45:


tzarebczan:
Thanks @ykris45 , good idea!

PR #817, opened 2018-10-10
fix backlink

ykris45:
screenshot_2018-10-11-05-45-32-491
i read the faq and i saw blabla its means image code . and it will error 404 on my device. but the better view is use baclink to the FAQ as mentioned on article
 [hacktobersfest]


tzarebczan:
Thanks Chris!

Issue #819, opened 2018-10-12
Remove quickstart


tzarebczan:
/quickstart will be redirected to lbry.tech/playground

ykris45:
i will do it . i think its so easy based on my experience.

alyssaoc:
closing this has been redirected

PR #844, opened 2018-10-20
bring back quickstart

ykris45:
@NetOperatorWibby bring back quickstart link and redirect to https://lbry.tech/playground... based on closes issues #819


tzarebczan:
We removed this on purpose, that section already leads to lbry.tech

PR #857, opened 2018-10-25
Removed unused variables in Session.class.php

gy741:
Hello,

In my opinion, No $val variables are required.

Because it is not used.

What do you think of it?

Thanks.


kauffj:
This will just result in iterating over the values.

You could iterate over `array_keys(xxx)` instead, but it's arguably worse to do this.

PR #858, opened 2018-10-26
updated office address

sridhareaswaran:


jsigwart:
@Feanor78 These contest rules were for previous contests and may or may not apply to future contests. Is the FAQ page the right place for this, or should each contest have their own rule page? Other ideas?

Feanor78:
@jsigwart the contest T&Cs can be a single page that we update based on what's running - there shouldn't be multiple terms and conditions.

tzarebczan:
@sridhareaswaran, thanks for the PR! Can we [send you some LBC as appreciation](https://lbry.io/faq/appreciation)?

PR #861, opened 2018-10-30
add mail helpdesk to footer

ykris45:
### Description

i added mail icon and redirect to hello@lbry.io when user or visitor clicked it via the mail icon on footer
image


tzarebczan:
Thanks @ykris45 !

PR #867, opened 2018-11-02
Update 190-october-community-update.md

ykris45:
fix telegram link , add some link email and telephone so reader can contact directly in this article. and also add category


tzarebczan:
Thanks for the PR @ykris45 !

PR #872, opened 2018-11-07
Move all image to Spee.ch

ykris45:
### Description
 some country in the world has blocked imgur image hosting so i think for now should move to spee.ch so users can access and view the image more better than before. and fix mistake for link


tzarebczan:
Thanks Chris, I like the initiative!

PR #882, opened 2018-11-15
Update 103-great-week-for-lbry-developer-beta.md

ykris45:
### Description

update links to newest version


tzarebczan:
Don't need to update these everywhere, that's what the permanent redirects are for. We'll eventually redirect the api docs too.

PR #884, opened 2018-11-16
Update CU.php

ceoger:
updated Tag from VIT to CU

### Description

updated Tag for CU

### Fixes #


ceoger:
closed, committed to #885

PR #886, opened 2018-11-18
Set unique user id for google analytics and facebook pixel.

tiger5226:


view/template/layout/_analytics_footer.php
```diff
@@ -8,13 +8,16 @@
...
+    ga('set', 'userId', '<?php echo $_SESSION[Session::USER_ID] ?>');
     ga('send', 'pageview');
 
    !function(f,b,e,v,n,t,s) {if(f.fbq)return;n=f.fbq=function(){n.callMethod? n.callMethod.apply(n,arguments):n.queue.push(arguments)}; if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0'; n.queue=[];t=b.createElement(e);t.async=!0; t.src=v;s=b.getElementsByTagName(e)[0]; s.parentNode.insertBefore(t,s)}(window,document,'script', 'https://connect.facebook.net/en_US/fbevents.js');
 
-   fbq('init', '1618717031725766');
+   fbq('init', '1618717031725766',{uid: '<?php echo $_SESSION[Session::USER_ID] ?>'}));
    fbq('track', '<?php echo Response::getFacebookPixelAnalyticsType() ?>');
 
+    <?php LBRY::logWebVisitor('PHP', $_SESSION[Session::USER_ID], $_SERVER['REMOTE_ADDR'])?>
```
kauffj: This technically breaks MVC, as a template shouldn't be initiating remote calls (unless it's doing so client-side).

There's not a super-obvious place to put this, I'd say either part of `Session::init` or in `Controller::execute` or `Controller::dispatch`.

kauffj: (An alternative design is to create one of an/or both `Actions::preExecute` and `Actions::postExecute`, to be called on both sides of `Controller::execute`)

kauffj:
@tiger5226 discussed this some with @lyoshenka and I think this does need to change a bit. Taking discussion to https://github.com/lbryio/internal-apis/pull/643

tiger5226:
@kauffj The changes for this are in and I tested it with the respective internal-apis branch as well. If you approve, I will merge when we merge the internal-apis stuff.

tiger5226:
@kauffj just flagging you again, since we will be merging the internal-apis supporting feature soon.

controller/Session.class.php
```diff
@@ -31,6 +33,9 @@ public static function init()
             static::setNamespace(static::NAMESPACE_DEFAULT, $oldSession);
         }
 
+        $response = LBRY::logWebVisitor('lbry.io', $_SESSION[static::USER_ID], $_SERVER['REMOTE_ADDR']);
```
kauffj: 1) Mildly better to be explicit about what we expect/want to be passed when `$_SESSION[static::USER_ID]` is not set.
2) lbry.io probably shouldn't be hardcoded and instead read from server

controller/Session.class.php
```diff
@@ -31,6 +33,9 @@ public static function init()
...
+        $_SESSION[static::USER_ID] = $response['data']['visitor_id'];
```
kauffj: As written, if this code ever breaks in a weird way (e.g. a typo in the above data check), would it end up creating a lot of users server side due to relying on empty session user id to create new users?

lib/thirdparty/LBRY.class.php
```diff
@@ -75,4 +75,9 @@ public static function editYouTube($status_token, $channel_name, $email, $sync_c
             return Curl::post(static::getApiUrl("/yt/update"), ['status_token' => $status_token, 'new_email' => $email, 'new_preferred_channel' => $channel_name, 'sync_consent' => $sync_consent], ['json_response' => true]);
         }
     }
+
+    public static function logWebVisitor($site, $visitorID, $IPAddress)
+    {
+        return Curl::post(static::getApiUrl("/visitor/new"), ['site' => $site, 'visitor_id' => $visitorID, 'ip_address' => $IPAddress], ['json_response' => true]);
```
kauffj: Consider how to handle this function only being called in production.

view/template/layout/_analytics_footer.php
```diff
@@ -8,13 +8,16 @@
...
+    ga('set', 'userId', '<?php echo $_SESSION[Session::USER_ID] ?>');
     ga('send', 'pageview');
 
    !function(f,b,e,v,n,t,s) {if(f.fbq)return;n=f.fbq=function(){n.callMethod? n.callMethod.apply(n,arguments):n.queue.push(arguments)}; if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0'; n.queue=[];t=b.createElement(e);t.async=!0; t.src=v;s=b.getElementsByTagName(e)[0]; s.parentNode.insertBefore(t,s)}(window,document,'script', 'https://connect.facebook.net/en_US/fbevents.js');
 
-   fbq('init', '1618717031725766');
+   fbq('init', '1618717031725766',{uid: '<?php echo $_SESSION[Session::USER_ID] ?>'}));
    fbq('track', '<?php echo Response::getFacebookPixelAnalyticsType() ?>');
 
+    <?php LBRY::logWebVisitor('lbry.io', $_SESSION[Session::USER_ID], $_SERVER['REMOTE_ADDR'])?>
```
kauffj: 1) This breaks MVC.
2) Isn't this redundant if it is in `Session::init()`? (Session::init is called on every execution, I believe)

tiger5226: You are right, init is called on every execution. I removed that line since it is already in the init.

tiger5226: it appears to get called twice for me...2 seconds apart.

controller/Session.class.php
tiger5226: done and done

controller/Session.class.php
tiger5226: So I tested this out, by replacing `visitor_id` with `viitor_id` and it actually causes the site to error for all users. So I added `key_exists` checks ( thanks @lyoshenka ) for these. Also you are right that it will endlessly create visitors on internal-apis. 

So I just made a change to link site visitors by ip address as well. If an ip address calls the api twice in 24 hours they will be regarded as the same site visitor. So at most, if this happens it will not create site visitors in internal-apis.

Are you ok with this approach?

lib/thirdparty/LBRY.class.php
tiger5226: I can add the `IS_PRODUCTION` check here and only return the curl post if that is true. I tested this out and it does work and stops the api from being called. What I am not sure of are the implications. If I return nothing, what does that mean in PHP? There are no errors in the console and the web page appears fine.

controller/Session.class.php
kauffj: Sounds sensible to me. Are you/we sure want this called on every request? If so, it might be sensible to do it after the response is flushed to the user (as-is this will presumably add 100+ms to each page render), but this doesn't need to block shipping.

lib/thirdparty/LBRY.class.php
kauffj: This function probably doesn't need to return at all, since the return is never checked or used anyway.

view/template/layout/_analytics_footer.php
kauffj: It's going to get called once on init (happens early in execution) as well as when this footer is rendered (towards the end of execution). I hope it doesn't take a full 2s to render a page though!

lib/thirdparty/LBRY.class.php
tiger5226: well it is used to store the visitor id in the php session object. This is how we give an id to the analytics.

view/template/layout/_analytics_footer.php
tiger5226: I removed it from the footer. So its somewhere else.

controller/Session.class.php
kauffj: I just checked the code and we currently do not do any output buffering.

So if you wanted to pursue above it'd look something like this.

- Add new method `Response::addPostRenderCallback(callable $func)`. This method stores these callables in internal Response array.
- Add new method `Response::invokePostRenderCallbacks()`, which invokes the callbacks.
- Update `Session::init()` to call `Response::addPostRenderCallback` wrapping the existing logic in an inline function.
- Update `index.php` to flush (I think one of and/or both `ob_flush()` or just `flush()`) and then call the above method.

tiger5226: I did what you said. Everything on Response is static. So I left it static. Is everything in PHP in the context of a single user?

kauffj: Yes.

lib/thirdparty/LBRY.class.php
kauffj: Right, my bad.

controller/Session.class.php
```diff
@@ -31,6 +34,18 @@ public static function init()
             static::setNamespace(static::NAMESPACE_DEFAULT, $oldSession);
         }
 
+        Response::addPostRenderCallback(function(){
+            $site_visitor_id = key_exists(static::USER_ID, $_SESSION) ? $_SESSION[static::USER_ID] : '';
```
kauffj: This is fine, but in "modern" PHP, a more common way to write it would be:

`$_SESSION[static::USER_ID] ?? ''`

`??` is a null coalesce operator.

kauffj:
LGTM but I did not actually run the code as I am not set up to do so on my backup/temp computer. 

Do you want me to merge anyway?

tiger5226:
I am going to merge the adjustments to internal-apis first. Then we can merge this.

controller/Session.class.php
tiger5226: Thanks! Keep these types of things coming. I am basically googling syntax for what I want to do and sometimes SO is out of date.

tiger5226:
merged and deployed

PR #888, opened 2018-11-19
add instant bitex

ykris45:
i hear this news from twitter and if Mr Thomas Zarebczsn retweet this tweet from instant bitex. its means as trusted exchange.. hehe


tzarebczan:
Moved to Regular exchange section since they require accounts.

ykris45:
@tzarebczan Thanks You sir. collaborate again with you

PR #897, opened 2018-12-05
fix error link and add redirect

ykris45:
i saw nginx error bad gateway and i change to open lbry


tzarebczan:
Thanks Chris!

ykris45:
you welcome sir

PR #899, opened 2018-12-06
Landing Page For Events We Attend

ceoger:
### Description
create this page so we can share about LBRY with people we meet in events

### Fixes #

### DIFF
--- a/view/template/page/CUniversity.php
+++ b/view/template/page/CUniversity.php
@@ -1,5 +1,5 @@
-<?php Response::setMetaDescription(__('LBRY Welcomes VIT Vellore\'s E-Cell HackerTech 2018')) ?>
-<?php Response::setMetaTitle(__('LBRY Welcomes VIT Hackathon Participants')) ?>
+<?php Response::setMetaDescription(__('LBRY Welcomes Chandigarh University Students')) ?>
+<?php Response::setMetaTitle(__('LBRY Welcomes Chandigarh University Students')) ?>
 <?php echo View::render('nav/_header', ['isDark' => true, 'isAbsolute' => true]) ?>
 <main>
   <div class="cover cover-dark cover-center cover-full cover--dark-overlay" style="background-image: url(/img/clement-vit.jpg)">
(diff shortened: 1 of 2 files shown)


view/template/page/events.php
```diff
@@ -0,0 +1,31 @@
...
+        <a href="#about" class="btn-primary btn-large">Tell Me More!</a>
+      </div>
+    </div>
+  </div>
+  <div class="cover cover-light" id="about">
+    <div class="content content-light content-readable">
+      <h3>Enter your email address to receive LBRY updates, news, and information for India's LBRYians!</h3>
+      <p>As an open source project, our development and future relies on our communities around the world, and we're delighted to see our community growing in India. Please enter your email below and we'll give you instructions to redeem the LBC as mentioned by our team.</p>
+      <?php echo View::render('mail/_subscribeForm', [
+        'tag' => 'events',
```
tzarebczan: Can we change this tag to something else? @ceoger  / @jsigwart . Maybe something more specific.

view/template/page/CUniversity.php
```diff
@@ -0,0 +1,31 @@
...
+        <a href="#about" class="btn-primary btn-large">Tell Me More!</a>
+      </div>
+    </div>
+  </div>
+  <div class="cover cover-light" id="about">
+    <div class="content content-light content-readable">
+      <h3>Enter your email address to receive LBRY updates, news, and information for India's LBRYians!</h3>
+      <p>As an open source project, our development and future relies on our communities around the world, and we're delighted to see our community growing in India. Please enter your email below and we'll give you instructions to redeem the LBC as mentioned by our team.</p>
+      <?php echo View::render('mail/_subscribeForm', [
+        'tag' => 'vit',
```
tzarebczan: new CU tag?

view/template/page/events.php
ceoger: not sure, this is not a specific event, there is an exchange launch happening in new delhi for which we plan to pass by, we may get to talk to the attendees about LBRY, so maybe we can use an existing landing page else please suggest a any?

view/template/page/CUniversity.php
tzarebczan: What do you mean by ignore? Do you still need this landing page?

view/template/page/events.php
tzarebczan: What's the purpose of tagging these people? Will you reach out to them? Maybe make a generic one like `india-other` for these one offs? Just didn't want a tag in our system called "events" - too generic.

view/template/page/CUniversity.php
ceoger: yes please, i still need it, the pr has previously been merged

tzarebczan:
I think we should also change the url to something more appropriate as well. Is the CU page no supposed to be merged? Might be better to create a new PR :)

PR #903, opened 2018-12-12
embed tweet

ykris45:
embedding tweet widgets


tzarebczan:
@ykris45 can you put this at the end instead?

ykris45:
yes i can. please wait at least 10 minutes from now

tzarebczan:
Thanks @ykris45 !

tzarebczan:
@ykris45 think I'm going to remove this since it just repeats what's already said in the blog and the same GIF.

Issue #559, opened 2018-05-16
Allow confirmation of email via lbry.io directly


kauffj:
A user should be able to confirm their email just by clicking a link, visiting lbry.io, and doing nothing else.

PR #907, opened 2018-12-14
feat: email auto verify page

seanyesmunt:
### Description

Depends on https://github.com/lbryio/internal-apis/pull/672, but I think this can be reviewed before that is merged. (The endpoint may change) 

This new page will replace the current email verification page (once all apps are using the new email verification flow). Not much is different, other than the verification now happens in the browser. We call `api.lbry.io` once a user completes the captcha.

Something I'm still unsure of is what should happen if it fails:
  - "refresh to try again"?
  - "go to your app and resend the verification email"?

### For #559

### DIFF
--- a/controller/action/AcquisitionActions.class.php
+++ b/controller/action/AcquisitionActions.class.php
@@ -27,7 +27,7 @@ class AcquisitionActions extends Actions
 
         return Controller::redirect(Request::getReferrer(), 303);
     }
-      
+
     public static function executeYouTube(string $version = '')
     {
         if (isset($_GET['error_message'])) {
@@ -57,6 +57,11 @@ class AcquisitionActions extends Actions
     public static function executeVerify(string $token)
     {
         return ['acquisition/verify', ['token' => $token]];
+    }
+
+    public static function executeAutoVerify()
+    {
+        return ['acquisition/auto-verify'];
     }
 
     public static function executeYoutubeToken()
--- a/view/template/acquisition/auto-verify.php
+++ b/view/template/acquisition/auto-verify.php
@@ -1,0 +1,71 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+    <title>Verify Your Identity</title>
+    <link rel="icon" href="images/favicon.ico">
+    <script src='https://www.google.com/recaptcha/api.js?' async defer></script>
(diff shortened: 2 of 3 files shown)


view/template/acquisition/auto-verify.php
```diff
@@ -0,0 +1,70 @@
...
+        <br/>
+        <div class="g-recaptcha" data-sitekey="6LcG_z0UAAAAAKBPDBhiJU_jI9cRNRiJwcUHq95u" data-callback="verifyCallback" data-expired-callback="expiredCallback"></div>
+    </div>
+
+    <div style="display: none; margin-top: 10px;" id="verify">
+        <p>Verifying...</p>
+    </div>
+    <div style="display: none; margin-top: 10px;" id="verify-error">
+        <p>There was an error</p>
+        <code id="verify-error-text"></code>
```
seanyesmunt: We should probably add other text here if an error occurs, I'm not sure what though.

kauffj: 1) Is there information from server that would be helpful to be displayed?
2) You can always encourage them to contact help@lbry.io

controller/action/AcquisitionActions.class.php
```diff
@@ -59,6 +59,11 @@ public static function executeVerify(string $token)
         return ['acquisition/verify', ['token' => $token]];
     }
 
+    public static function executeAutoVerify()
```
kauffj: FYI, you could have omitted a route definition and controller function entirely if you had just put the template as `page/verify` (there is a catch all route for this pattern to avoid writing empty controllers). However, in this case, since one would reasonably expect a controller to exist, I think it's fine/good to have the route defined.

view/template/acquisition/auto-verify.php
```diff
@@ -0,0 +1,70 @@
...
+            document.getElementById("verify").style.display = "block";
+        }
+
+        var expiredCallback = function(error) {
+            console.log("expired", error)
+        }
+    </script>
+</head>
+<body>
+<div style="display: flex; align-items: center; flex-direction: column;" class="text-center">
```
kauffj: I didn't load these styles but they look copied from the previous experience. 

@NetOperatorWibby this is another decent candidate for a first lbry.io style to try unifying via components, since it is cleanly isolated from the rest of the site and it'd make sense to match the app UI aesthetic (you can only end up here from the app).

Not a blocker to merging/deploying.

view/template/acquisition/auto-verify.php
seanyesmunt: Added a comment about contacting help@lbry.io if an error keeps happening.

PR #920, opened 2019-01-13
Fixed URL for Simple Swap

MyNameIsJulian:
### Description

Currently the entry for SimpleSwap has www. in front of it. The site doesn't work with www. in front of it because of an SSL misconfiguration on their end.

### Fixes #

Removed www.


tzarebczan:
Thanks @Julian2000nl !! Can we [show you some appreciation](https://lbry.io/faq/appreciation) for fixing this up?

Julian2000nl:
@tzarebczan Of course! I've send an e-mail to the address on the page. :smile:

PR #938, opened 2019-01-28
Redesign

NetOperatorWibby:
### Description

[A whole new world](https://www.youtube.com/watch?v=hZ1Rb9hC4JY)

### Fixes

E V E R Y T H I N G


web/scss/partials/_header.scss
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+.header__navigation-children {
+  display: flex;
+  flex-wrap: wrap;
+  position: relative;
+
+  li {
+    &:not(:last-of-type) {
```
kauffj: Not necessary to fix and possibly worth a discussion in chat with app team generally on pros/cons, but I've found `li + li { margin-top: 1rem; }` to be a mildly preferable approach.

view/template/nav/_footer.php
```diff
@@ -1,36 +1,51 @@
...
-            <img src="/img/Free-speech-flag.svg" alt="Free Speech Flag" height="30"/>
-          </a>
-        </div>
-      </nav>
-      <div>
-        <a href="/join-us">LBRY is hiring!</a>
-      </div>
-    </div>
+<footer class="footer">
+  <div class="inner-wrap">
```
kauffj: Why use `.inner-wrap` repeatedly inside of elements instead of just naming this `.footer__inner`? 

(Analogous question for several other areas.)

web/scss/pages/_ancillary.scss
```diff
@@ -0,0 +1,219 @@
...
+  blockquote,
+  iframe,
+  ol,
+  p,
+  ul {
+    margin-bottom: 1.25rem;
+  }
+
+  ol,
+  p,
```
kauffj: Do we need unique content stylings for our ancillary pages vs. blog posts or other more mainstream pages? Same for tables and probably some other rules in this sheet.

web/scss/_layout.scss
```diff
@@ -0,0 +1,141 @@
...
+  font-size: 12px;
+}
+
+
+
+code {
+  @include font-mono;
+}
+
+h1,
```
kauffj: Is there a shared scope it would make sense for some of these rules (particularly the header stylings) can go in rather than being global? I suspect a number of custom pages or elements would want to present an h3 differently.

web/scss/pages/_ancillary.scss
NetOperatorWibby: Everything in this file will most likely be moved to the "rich text" component. I gotta look at how the app is doing things and basically pick and choose. Or, just finish this site and make the app use the same thing.

view/template/nav/_footer.php
NetOperatorWibby: This enables us to have divs (like the footer) to have a background that extends the width of your screen, while also keeping a container. `.inner-wrap` is a global class so that's why you see it everywhere.

web/scss/partials/_header.scss
NetOperatorWibby: `margin-top` is not preferable to me because the flow of content is top-down, not down-up. In certain circumstances I'll use it but it doesn't need to be here.

web/scss/_layout.scss
NetOperatorWibby: That makes sense, I could move these to `.ancillary` for the time being.

web/scss/pages/_home.scss
```diff
@@ -0,0 +1,123 @@
...
+    font-size: 2rem;
+    margin-bottom: 1rem;
+  }
+
+  p {
+    font-size: 1.25rem;
+    line-height: 1.33;
+    margin-bottom: 2rem;
+  }
+}
```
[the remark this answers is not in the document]
NetOperatorWibby: 🙌🏾

web/scss/components/icons/_a.scss
```diff
@@ -0,0 +1,157 @@
+// A
```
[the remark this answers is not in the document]
NetOperatorWibby: They were in here and I assumed you (or someone else) knew.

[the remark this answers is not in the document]
NetOperatorWibby: Personally, I'd rather use an SVG sprite.

NetOperatorWibby:
@kauffj PR is (mostly) complete. I think the change of `lbry.io` to `lbry.com` should be a PR after this one is merged. Or I can make a PR based on this one.

NetOperatorWibby:
@kauffj I haven't done anything new to the CSS since our BEM convo last week. I know what needs to be done.

PR #940, opened 2019-02-01
Splash pages for our upcoming sites

NetOperatorWibby:
### Description

Title says all. New routes created are `/org` and `/tv`.

### Fixes

The terrible absence of these pages are now filled.

### DIFF
--- a/controller/Controller.class.php
+++ b/controller/Controller.class.php
@@ -79,6 +79,11 @@ class Controller
 
         $router->get(['/', 'home'], 'ContentActions::executeHome');
 
+        //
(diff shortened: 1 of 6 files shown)


kauffj:
@NetOperatorWibby looks good! I did change the routing, you can see how I did it here: https://github.com/lbryio/lbry.io/pull/940/commits/ec3fb40f92eec396dfa9de865009d7ab2b702587

I'm actually not sure if this is correct either. We'll see once https://github.com/lbryio/ansible/issues/10 is completed.

Will go over these changes with Rob on Monday, as well as copy.

PR #947, opened 2019-02-05
Fix typo in example settings file (peer_port rather than peer:port)

goldenapple3:
### Description
Noticed a typo in the [How to change default daemon port?](https://lbry.io/faq/how-to-change-port) article. The example `daemon_settings.yml` file it provides is incorrect: 

    {download_directory: c:\users\lbry,
    peer:port: 3334}

It should say `peer_port` rather than `peer:port`. A bit of an annoying typo considering LBRY fails to start because of it. :)

### DIFF
--- a/content/faq/how-to-change-port.md
+++ b/content/faq/how-to-change-port.md
@@ -12,7 +12,7 @@ The most user friendly way to change the port permanently is to append the below
 Sample daemon_settings.yml (may vary by OS):   
 
     {download_directory: c:\users\lbry,
-    peer:port: 3334}
+    peer_port: 3334}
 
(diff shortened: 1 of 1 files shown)


tzarebczan:
Nice catch and thanks for fixing it! Can we [show you some appreication](https://lbry.io/faq/appreciation) for your time?

goldenapple3:
I'm not sure if fixing a typo is really a major contribution, but I'll take the credits, if you don't mind.

E-mail sent.

tzarebczan:
Sent, thanks again! Every bit of help counts, thanks for checking out LBRY!!

PR #948, opened 2019-02-06
2019 roadmap

kauffj:
This updates our roadmap to pull from the internal issues repo and show all key initiatives tagged 2019.

This PR also:

- Removes the display of past release notes from the roadmap (@tzarebczan, what do you think of this?)
- Switches GitHub authentication to use a personal auth token rather than an application token (@lyoshenka what do you think of this? more context [here](https://github.community/t5/GitHub-API-Development-and/How-do-I-grant-an-application-access-to-read-Issues-from-a/m-p/18546#M839))

Due to the GitHub authentication change, this PR CANNOT BE MERGED until @nikooo777 updates Ansible server config to store the token. This also means it'll be tricky to develop the roadmap locally, though that's easily fudge-able by committing a sample API response to be used in dev environments.


NetOperatorWibby:
So what would I have to do for development?

web/scss/_roadmap.scss
```diff
@@ -117,6 +117,17 @@ $title-spur: 8 * $width-roadmap;
 {
   font-size: 0.9em;
   margin: 10px;
+  /* Paul, when you rebase on master and see this, these are temp rules I added to help the roadmap look only like lukewarm garbage rather than hot garbage*/
```
NetOperatorWibby: 👍

kauffj:
@NetOperatorWibby I'll add sample data for development purposes

PR #968, opened 2019-02-23
Added information about close to system tray

JohnBlood:
When I tried to restore my backed up data, it didn't work until someone mentioned the fact that LBRY minimizes to the system tray when ti tell it to close. Decided to add it here to help others.

### Fixes #


tzarebczan:
Thanks for the PR John! Will send you some appreciation on Discord. We'll also be adding this option to the desktop app (enable/disable)

PR #977, opened 2019-03-07
Case insensitive URL

### LINKED ISSUE #556: Drop case sensitivity in URLs

maximest-pierre:
### Description
case insensitive URL

### Fixes #556

### DIFF
--- a/controller/Controller.class.php
+++ b/controller/Controller.class.php
@@ -59,7 +59,7 @@ class Controller
             $dispatcher = new Routing\Dispatcher($router->getData());
             return $dispatcher->dispatch($method, $uri);
         } catch (\Routing\HttpRouteNotFoundException $e) {
-            return NavActions::execute404();
+            return static::doLowerUriRedirectOr404($uri);
         } catch (\Routing\HttpMethodNotAllowedException $e) {
             Response::setStatus(405);
             Response::setHeader('Allow', implode(', ', $e->getAllowedMethods()));
@@ -163,7 +163,7 @@ class Controller
                 Response::enableHttpCache();
                 return ['page/' . $slug, []];
             } else {
-                return NavActions::execute404();
+                return static::doLowerUriRedirectOr404($slug);
             }
         });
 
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for the contribution @maximest-pierre!!! We'll get someone to review this. Have you tested it locally? Can we send you some [appreciation](https://lbry.io/faq/appreciation) for your effort?

maximest-pierre:
Yes, everything was tested locally with dynamic routes and static routes.

NetOperatorWibby:
@maximest-pierre I'll review this and your other PR in the AM!

NetOperatorWibby:
@maximest-pierre I just copied/pasted your code into `controller/Controller.class.php` and navigating to `localhost:8000/NEWS` gave me a 404 error.

controller/Controller.class.php
```diff
@@ -198,4 +198,29 @@ public static function shutdown()
             call_user_func($fn);
         }
     }
+
+    protected static function doLowerUriRedirectOr404($uri)
+    {
+        $router = static::getRouterWithRoutes();
+        // $lowerUri = strtolower($uri);
```
NetOperatorWibby: Comments shouldn't be merged.

controller/Controller.class.php
```diff
@@ -198,4 +198,29 @@ public static function shutdown()
...
+        $lowerUri = '/news/lbry-evolves';
```
NetOperatorWibby: Why does this function specify this URL?

maximest-pierre: I hard coded it during one of the test.

NetOperatorWibby: Tests shouldn't be in the final code.

controller/Controller.class.php
```diff
@@ -198,4 +198,29 @@ public static function shutdown()
...
+                return static::redirect($lowerUri, 308);
+            }
+        }
+        return NavActions::execute404();
+    }
+
+    protected static function checkForRoute($uri, $router)
+    {
+        $routerData = $router->getData();
+        if (in_array($uri, $routerData->getStaticRoutes())) {
```
kauffj: 1) If this code was valid, it could simply be: `return in_array(xxx) || in_array(zzz)`
2) This code is not valid. You have to perform the actual routing to see if the URI matches, you can't simply do a key check.

PR #979, opened 2019-03-08
partial nav changes

kauffj:
I plan to look at this again, some are still missing links and need better copy

### Fixes #


kauffj:
@NetOperatorWibby I'm done with changes here, though there's still no contact page. A simple contact page would also solve the email in the footer issue.

If there are still formatting issues I recommend finding/adding a deterministic formatter.

PR #980, opened 2019-03-10
Update vit.php

ceoger:
updated the file for womentechies hack

### Fixes #


ceoger:
@jsigwart can you have a look at this for me

PR #982, opened 2019-03-18
Add Dev Program +  Forum as drawer items

tzarebczan:
Thoughts?

### Fixes #


NetOperatorWibby:
Jeremy didn't want these in the nav so I removed them a while ago.

PR #1005, opened 2019-03-26
add GPS Link to the Office Address

ykris45:
### Description
for make visitor know where LBRY Office Located and also if they click this link it will open google maps apps and setting up GPS Navigation from where visitor lived to make LBRY Journey easily
image


tzarebczan:
Thanks @ykris45 !!

PR #1014, opened 2019-03-30
Update tipbot-twitter.md

ykris45:
### Description

image

alternative to fix the embedding tweet was dissapear from this FAQ,


tzarebczan:
So you removed the tweet objects and replaced with images?

ykris45:
yes thats true, its same as before, because as paul said on last issues i was open for LBRY Inc

tzarebczan:
Thanks @ykris45 !

PR #1017, opened 2019-04-02
Added Icons to Footer

ykris45:
### Description

Would Be Nice if LBRY.com Website Footerr have a icon for each menu , it will make it looks badass
image


tzarebczan:
I like the initiative @ykris45 ! We'll let Paul take a look to see if this is what he wanted, or wanted to keep it clean with only the titles.

NetOperatorWibby:
I don't think icons are necessary here.

PR #1024, opened 2019-04-26
fix incorrect email address

ykris45:
### Description

change lbryio to lbry.com and added . to this email


tzarebczan:
Thanks @ykris45 !

PR #1025, opened 2019-04-26
Edits about publishing

eggplantbren:
I've noticed some users with incomplete uploads to the reflector, so I've made a couple of edits to the publishing FAQ to hopefully emphasise that they should leave the app open for a while even after the publish _to the blockchain_ is complete.

I also edited the thing about how to tell if someone is viewing your content, because there are view counts now.


tzarebczan:
Thanks @eggplantbren !!

PR #1048, opened 2019-05-25
Ignore timeout curl exceptions

voidfire:
### Description

Throw Curl exceptions Unless if they are timeout related.

As far as I can see all exceptions generated are being forwarded to slack thus I ignored these specific exceptions from Curl.class.php

Another solution would be to change the hardcoded  5 seconds timeout limit on curl.class.php.
_This bit of functionality is **hard to test from my side** with no slack access, but it should filter properly those exceptions on curl._ 
### Fixes ##1023

### DIFF
--- a/lib/tools/Curl.class.php
+++ b/lib/tools/Curl.class.php
@@ -144,7 +144,10 @@ class Curl
                 $options['retry'] -= 1;
                 return static::doCurl($method, $url, $params, $options);
             }
-            throw new CurlException($ch);
+            // Ignore the timeout exceptions
+            if (!in_array(curl_errno($ch), array(CURLE_OPERATION_TIMEDOUT, CURLE_OPERATION_TIMEOUTED))) {
+                throw new CurlException($ch);
+            }
         }
 
         curl_close($ch);


tzarebczan:
Thanks for the PR @voidfire , we'll get this reviewed! Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the help?

kauffj:
Apologies for the Slack logging. While this project is open source like everything at LBRY, this project is not as commonly developed by others and not as user-friendly. If you wanted to add support for multiple logging methods configurable by `config.php`, we'd be happy to accept that change.

I'm unclear on why timeouts should not except, as it's ambiguous what is expected when a `cURL` times out. If you're looking to differentiate between timeouts and other failures, how about adding a different named exception for the timeout case?

kauffj:
(I read the issue after commenting. I'm a jackass.)

I understand why now (some calls to log user visits timeout), but I think it would still be better to explicitly handle this case then simply ignore them.

kauffj:
@voidfire @tzarebczan I'm going to close this. Still happy to merge/fix this if we can instead handle it via a specific timeout exception that can be ignored in the cases it ought to be.

PR #1053, opened 2019-06-19
Change /faq/claimtrie-implementation to a link

J0hnL0cke:
Changing this URL to a link allows users to click it instead of having to edit the URL in their browser, which may cause some confusion. The target location (lbry.com/faq/claimtrie-implementation) is still shown as plain text, so the user knows where they are being redirected to.


tzarebczan:
Hey Josh, thanks for the contribution! Can we show you some [appreciation](https://lbry.com/faq/appreciation)

tzarebczan:
I've updated this on master with the correct link.

PR #1064, opened 2019-06-27
Update exchanges.md

Draiderz:
### Fixes #


tzarebczan:
Is it not available there? I can't register from the US.

Draiderz:
In the site does not exist anything about LBRY, nor to wallet or a marketplace, and at instantbitex it is not possible to deposit, they do not provide an address.

Draiderz:
> Is it not available there? I can't register from the US.


In the site does not exist anything about lbry, nor to wallet or a marketplace, and at instantbitex it is not possible to deposit, they do not provide an address

PR #1069, opened 2019-07-10
Update githubroadmap.php

jsigwart:
fixes growth#193

### Fixes #


kauffj:
@jsigwart the file you located is dummy data only used for local development when you might not have a GitHub API credential.

The live roadmap items are pulled from these tickets: https://github.com/lbryio/internal-issues/issues?q=is%3Aissue+is%3Aopen+label%3A2019

If you edit one of those tickets, the change should be live within an hour (the website caches results from GitHub).

PR #1073, opened 2019-07-24
Do Not Merge : Update Screenshot on Publish FAQ

ykris45:
i will finish it tommorow.. this is my entry,,hehehe


tzarebczan:
Thanks Chris,  but @ceoger took a closer look at this one and removed some of the more obvious screenshots.

PR #1112, opened 2019-09-05
Updated preview video

NetOperatorWibby:
'nuff said

### DIFF
--- a/view/template/download/get.php
+++ b/view/template/download/get.php
@@ -49,8 +49,8 @@
             playsinline
             style="display: block; margin-right: auto; margin-left: auto; max-width: 700px; "
           >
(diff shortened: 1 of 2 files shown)


tzarebczan:
Thanks Paul. Seems a bit too fast - can you try 2.5 seconds...that's what I've used before (should have mentioned this).

Also, Chrome won't play the current one on lbry.com for me at time...I have to refresh a bunch. Any ideas?

NetOperatorWibby:
Chrome probably has the original video cached. I'll update when I get to my desktop again.

NetOperatorWibby:
Updated the video

tzarebczan:
Thanks, looks good! Let's see how it looks on lbry.com now.

PR #1114, opened 2019-09-06
Closes #1109

### LINKED ISSUE #1109: Display bug when verifying email subscription
image

to repro, add a new email like tom+testdev94@lbry.com on lbry.tech, and then follow confirmation email.

NetOperatorWibby:
Confirmation message is no longer in the corner. At some point a more elegant solution will replace these inline overrides.


NetOperatorWibby:
Closes #1109

PR #1122, opened 2019-09-21
maybe it should help og image not showing on Tweet

ykris45:
https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary-card-with-large-image my references


kauffj:
@ykris45 those are already on the page, but thanks for trying!

PR #1123, opened 2019-09-24
Removed homepage exception for twitter image (fix for twitter image n…

### LINKED ISSUE #1110: No OG image on Twitter
See this thread: https://twitter.com/TomZarebczan/status/1169359128894070784

Missing images for FAQ artcles and /get link.

/Android shows the below - I was still expecting the main og image if not on Android. 
image

adnan360:
…ot showing)

### Description

Removed a homepage exception that was stopping from the Twitter image to not show on homepage (lbry.com). Image was not showing on homepage (lbry.com) but shows on other URL (lbry.com?test=123). It was because "content" attribute for a "twitter:card" meta was being set to "app" for homepage which was causing the image to not show for the home url. Submitted in response to: https://twitter.com/jeremykauffman/status/1174725053877039104
Solution was proposed in my tweet: https://twitter.com/adnan360/status/1176393294030364672

### Fixes #
Fixes #1110 Fix for Twitter card image not showing for homepage

### DIFF
--- a/view/template/layout/basic.php
+++ b/view/template/layout/basic.php
@@ -63,7 +63,7 @@
         <meta name="twitter:image" content="<?php echo $url ?>" />
     <?php endif ?>
 
-    <?php if (in_array($_SERVER['REQUEST_URI'], ['/', '/android', '/get'])): ?>
+    <?php if (in_array($_SERVER['REQUEST_URI'], ['/android', '/get'])): ?>
       <meta name="twitter:card" content="app"/>
       <meta name="twitter:app:country" content="US"/>
       <meta name="twitter:app:name:googleplay" content="LBRY beta"/>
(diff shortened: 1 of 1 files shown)


kauffj:
@adnan360 I'm uncertain about this change. The primary purpose of lbry.com homepage is to facilitate app installs, so we thought it made sense for links to lbry.com to show the app card.

Do you know if there is any way to have the app card show and still get a working fallback for desktop and/or non-mobile users?

adnan360:
> Do you know if there is any way to have the app card show and still get a working fallback for desktop and/or non-mobile users?

I think we would have to choose one or the other. The "content" attribute can be either image or app.

We cannot even detect any twitter bot or useragent to show appropriate card for desktop / mobile users because it is cached on Twitter's end. Once it is cached, we cannot change it depending on the device. If we want to show both of them, Twitter themselves will have to implement this feature and maybe even a new rules for having both of them in HTML metas.

Issue #1139, opened 2019-10-03
navigation drawer does not work


mkan0141:
#### Issue Summary

The navigation drawer does not work like the following GIF

![w1dv8-vjojt](https://user-images.githubusercontent.com/26357644/66110284-7394d400-e601-11e9-91e6-aa7c61f0822a.gif)

#### Steps to Reproduce

Reduce to mobile device size.

 I would like to work on this as a part of Hacktober Fest !!

tzarebczan:
Nice find @mkan0141 ! This used to work, so it must have gotten broken recently. Adding the label for you. Thanks for looking into it!

kauffj:
@mkan0141 I'm having trouble reproducing issues with the menu. It appears to work in production, both on an actual Android device and in Chrome on Desktop set to a mobile view.

Ah, I bet this was also caused by the JS bug introduced in https://github.com/lbryio/lbry.com/commit/b181563aba3728e212ebeca2b7d08fd5745af6c6...

@mkan0141 I'm going to close this issue and PR, but if you'd still like some [appreciation](https://lbry.com/faq/appreciation) for this please follow instructions in the link.

PR #1141, opened 2019-10-03
fix navigation drawer

mkan0141:
Fixes issue #1139 

### Description

The navigation drawer didn't work, so I solved it using the jQuery "toggleClass"

### Fixes #

before
w1dv8-vjojt

after
mx6bw-4ajb1


kauffj:
https://github.com/lbryio/lbry.com/issues/1139#issuecomment-537983835

Issue #1131, opened 2019-09-30
Words cut-off on window resize


deutranium:
#### Issue Summary

The navbar text cutoffs on resizing the window as illustrated in the following screenshot:

![Screenshot from 2019-09-30 23-10-44](https://user-images.githubusercontent.com/29124655/65902873-f71fac80-e3d8-11e9-97b7-7ad39e98e009.png)

#### Steps to Reproduce

Resize the window horizontally, and notice how the text wraps to the next line just before switching to the mobile-like view

If it's fine, then I would like to work on this as a part of Hacktober Fest

jsigwart:
I have added the Hacktoberfest label for you. :)

kauffj:
Great catch @deutranium, this is entirely appropriate!

liloo2040:
Hello! May I help you on this issue ?

kauffj:
@liloo2040 and/or @deutranium either you are welcome to work on this, though I just tried and could not reproduce it. It is possible this was a bug that was temporarily live for only a few days when we accidentally deployed some broken JS.

deutranium:
@kauffj I can still reproduce the bug, but I am not able to reflect the changes made in web/components/sass/navigation/_index.scss in the webpage

liloo2040:
Ok, I've seen the issue. I'm working on it.

ykris45:
@deutranium how to reproduce and notice on how the text wrapped in to mobile view?. i still not found this bug on my local machine?

tzarebczan:
This has been fixed.

PR #1147, opened 2019-10-07
word fix on navigation

liloo2040:
Issue #1131 : Words were cut while resizing the navigation bar to mobile.

Fix : I have done a word-wrap fix on the right file and the word-cutting no longer exists.

### DIFF
--- a/web/scss/_layout.scss
+++ b/web/scss/_layout.scss
@@ -1,6 +1,7 @@
 html,
 body {
-  width: 100%; height: 100%;
+  width: 100%;
+  height: 100%;
 }
 
 html {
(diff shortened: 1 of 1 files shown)


kauffj:
@liloo2040 there are a few problems with this fix:
1. `overflow-wrap` is the standardized name for this property, `word-wrap` is a legacy property.
1. Much more importantly, it doesn't appear to do anything. The `overflow-wrap` property of that element is already `normal`. You can verify with this code in the browser console: `window.getComputedStyle(document.getElementsByClassName('inner-wrap')[0], null).getPropertyValue("word-wrap")`
1. If this were the proper fix, it would be better to scope it more narrowly rather than applying to `.inner-wrap` which appears on many pages.

PR #1154, opened 2019-10-11
included og:url (#350)

### LINKED ISSUE #350: SEO and social metadata clean up (rel=canonical and og:url)
Two related issues:

### Missing `rel=canonical` on identical pages

@tzarebczan inadvertently pointed out that the following URLs serve the same post:

https://lbry.io/news/lbry-development-community-update-jan-2018
https://lbry.io/news/159-lbry-development-community-update-Jan-2018

At a minimum, blog post pages should include a `rel=canonical` tag pointing to the best URL (just the slug, no numbers, all lower case).

In addition, alternative URLs could be redirected to the preferred URL.

### Include og:url

Add an og:url tag that matches the `rel=canonical` URL, or the URL of the accessed page if there is no canonical URL

mevrouwdevries:
### Description

Included php code to extract and populate the og:url

### Fixes #350

### DIFF
--- a/view/template/layout/basic.php
+++ b/view/template/layout/basic.php
@@ -47,6 +47,10 @@
     <meta property="og:site_name" content="LBRY"/>
     <meta property="og:title" content="<?php echo $title ?>"/>
     <meta property="og:type" content="article"/>
+    <?php $url = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on' ?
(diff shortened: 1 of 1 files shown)


kauffj:
I'm not sure there is much if any benefit to simply always setting `og:url` to the requested URL (feel free to provide a resource or reference that suggests otherwise). I believe the purpose is to use this value when multiple URLs can result in the same content, which sometimes does happen on lbry.com, and this does not address.

PR #1168, opened 2019-11-01
fix missing image on twitter tipbot FAQ

ykris45:
i see on this faq recently and other images was lost so i fix that with full of love


tzarebczan:
Thanks for the PRs Chris!

PR #1181, opened 2019-11-15
fix url for this news

ykris45:
### Description
fix url

### Fixes #


ykris45:
@tzarebczan i fix this invalid link for moreno lbry blognews

PR #1204, opened 2020-01-10
Fixed typo in Where is my content stored and shared from

joebass85:
### Fixes #
#1002 Fixed typo in Where is my content stored and shared from


tzarebczan:

PR #1209, opened 2020-01-21
hacky app image support

kauffj:
this:

- makes all "download for linux" buttons link to /linux rather than specific files unless `preferredExt` is passed specifically
- makes /linux show buttons for appimage and deb, if they exist
- makes `/release/<repo>.ext` URLs work if repo has multiple release types for an OS

It is done in a semi-hacky way. If we ever needed to add a third Linux release type, it'd be easy, but if we ever needed to support multiple build artificats on another OS, it should be more fundamentally overhauled.


kauffj:
@tzarebczan not really expecting a thorough review, but giving you merge control

PR #1210, opened 2020-01-23
Corrected string from GPG to PGP

TigerxWood:
### Fixes #


kauffj:
🤦

Email help@lbry.com with an LBC address and a link to this comment if you'd like a thank you.

PR #1222, opened 2020-03-02
Update lbry-basics.md

PeterMortensen:
Some copy editing (e.g. ref. <https://en.wiktionary.org/wiki/access> - both as noun and verb), including elimination of run-on sentences.


tzarebczan:
Thanks for the PR! Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

PR #1223, opened 2020-03-04
Add Swapzone (Instant Exchange Aggregator)

SwapzoneIO:
### Description

 Swapzone aggregates exchange orders from SimpleSwap and StealthEX, so it definetly should be in the list.

### Fixes #


kauffj:
@SwapzoneIO merged, thank you. If anyone from your team needs to be notified re: hardforks or potential blockchain disruptions, you can subscribe to a list here: https://lbry.com/forklist

PR #1229, opened 2020-03-18
fix error 404 link

ykris45:
### Description

image


tzarebczan:
Nice catch, thank you!

PR #1253, opened 2020-04-30
Fix poor translation

mbrixius:
Due to high demand, you will be only be synced if you have at least 100 subscribers is correct english

### Fixes #


tzarebczan:
Thank you, will reach out via email to show some appreciation!

Issue #1252, opened 2020-04-28
Fix meta image on Twitter for lbry.com/get


kauffj:
https://twitter.com/LBRYio/status/1255248879815188481

kauffj:
Either the page for no system or Android is messed up

littlemadhatter:
Took a look at this, below are meta tags you should be able to incorporate into the site for it to pull correct meta tags, readable more at https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary

```
<meta name="twitter:card" content="summary" /> (Do not change)
<meta name="twitter:site" content="@LBRYio" /> (not required, though)
<meta name="twitter:title" content="LBRY, a decentralized content platform" />
<meta name="twitter:description" content="LBRY is a content platform owned by everyone (and no one)." />
<meta name="twitter:image" content="link" /> (read more about image at the link beforehand)
```
You can't test beforehand until the URL is live unless I skipped that part

PR #1255, opened 2020-05-02
fix meta image on twitter card

ykris45:
### Description
to fix issues no #1252. this script is not include meta image twitter property so i add it in to this script with this thumbnail LBRY default https://spee.ch/@thumbnails:4c/b8317a4f72e8cd64.jpeg


tzarebczan:
Don't think that's the image we want and don't think that's the right fix. There's already an og image and it's not showing .

ykris45:
@tzarebczan 
image see my works on LBRY Fund repo. twitter image metadata is different with og image property

ykris45:
@kauffj could you see my changes sir?

PR #1256, opened 2020-05-03
Correct an error

eatdostacos:
The title says it all.

### Description

I made a word the correct word. (Also, someone should check out the pictures and add new ones. They have apparently been taken off of spee.ch so they don't show up. I will create an issue.)

### Fixes #


tzarebczan:
Thanks, I fixed the images too!

PR #1264, opened 2020-05-06
grammer update to lbry-basics.md 

hibyemy:
I spend about 3 hours looking through spelling, grammar, and word choice to ensure that lbry would have the best presentation. 

There are a few word choice that I didn't fix because my goal is to make it Grammarly correct without changing what the author might have meant to say.

This page should be completely free of grammar errors except that few that I left out like for example: "check if in the browser there is no active plug-in or extension that act like a VPN" it should act as a VPN or acts like a VPN.


kauffj:
These looked good, thanks! Please email hello@lbry.com with a link to this pull request if you'd like an LBC tip @hibyemy

hibyemy:
singular of data is datum or something like that

hibyemy:
but data can be used as a singular

PR #1269, opened 2020-05-10
lots of small change to android page 

hibyemy:
I was only able to do a single page today as its one of the longer ones but did some corrections to grammar and also one of the image wasn't showing so I also fixed that.

![Choose an option]https://spee.ch/4/chooses.png)
Choose an option

^^^ that one


hibyemy:
oops sorry I accidentally included the last commit i did on may 5th

PR #1273, opened 2020-05-10
Change past info to past tense

eatdostacos:
I changed some of the parts that already happened to past tense to make it easier for new people to understand that that already happened.

### Fixes #


tzarebczan:
Thanks again!

PR #1280, opened 2020-05-12
Created help in FAQ page for changing blob files folder

### LINKED ISSUE #931: Create FAQ for moving data directory (+blobs)
#### Issue Summary
Create step by step FAQ on how to move your data_directory (takes up lost of space due to blob files). This was requested by a couple of community members. 

I think thill will eventually make it into an app side setting, but want to address in the near term.

TigerxWood:
### Description

Created help in FAQ page for changing blob files folder

Closes #931


tzarebczan:
Thank you so much for this @TigerxWood ! I'll give it a review soon. I did come across a Windows app that makes this process easier by moving the files/doing it automatically, so this might be a better option for some users - https://www.ghacks.net/2010/03/26/symbolic-link-creator/

Can we show you some [appreciation for this](https://lbry.com/faq/appreciation)?

TigerxWood:
Well I know there are these type of programs, also in MacOS there is a plug-in for Finder, but I thought that is much easy to wrote 2 or 3 commands and it's over. After all, this operation is make it only once, so I don't see the need to make the user to download other tools. But as you said, maybe for some these tools are usefull if they want to use them in more other occasions. So, I will modify the tutorial with these tools for windows and mac and make a new pull request.

TigerxWood:
I have updated the file with info using a graphical tool in Windows. In MacOS are too many steps to install a plug-in in Finder and then make a simlink, while with just 3 commands you can do the job in Terminal. So I think this is enough for the moment.

tzarebczan:
@TigerxWood made some changes - can you also replace the deletion step and instead have them copy the files over. I don't think we want users deleting the files if they plan to copy them over somewhere, especially if they are publishers.

tzarebczan:
Thanks again! Don't forget to reach out about the appreciation!

PR #1283, opened 2020-05-14
Fix links

acu192:
### Description

Fix one broken link (for Luxor) and one link without a protocol (for Blazepool, it was acting like a relative link oops).


tzarebczan:

acu192:
@tzarebczan Thanks for merging, and for the link. I sent in an email as instructed.

PR #1285, opened 2020-05-15
Update examples to working ones

keikari:
Original lines didn't work for me. But api call found in documentation seems to do wanted thing. Cli command I figured out with few tries. Tested with latest LBRY downloaded from AUR.

### Fixes #


tzarebczan:

PR #1294, opened 2020-05-22
Some more changes

eatdostacos:
### Description

I deleted a coming soon page for lbry.tv and made the page redirect to lbry.tv. I also add two more reidrects to the block explorer.

### Fixes #


tzarebczan:
Thank you!

PR #1297, opened 2020-05-24
Grammar1

hibyemy:
I committed the correct thing to my master and the wrong thing to the branch grammar1 sorry about that. So here is the correction.


tzarebczan:
Conflicts, can you repull from master and fix? Thank you!

PR #1299, opened 2020-05-25
small change 

hibyemy:
Had an hour-long debate with eatdostacos on if "We continue to hold the philosophy that focuses on building, both in tech and usage, is the only trend that matters in the long term." is correct or "We continue to hold the philosophy that focusing on building, both in tech and usage, is the only trend that matters in the long term."


tzarebczan:
Thank you for the PR!

PR #1303, opened 2020-05-30
dmca.md - fix typo and sentence structure.

infinite-persistence:
### Description
1. I don't think `legitimary` exists in the dictionary, so I changed to `legitimacy`. But I could be wrong.
2. Fixed multiple `also` in a particular sentence.

### Fixes #
N/A


tzarebczan:
Thanks for the PR!

PR #1308, opened 2020-06-04
Very Important Slash

eatdostacos:
When I clicked the link it said page not found because this slash was not there.

### Fixes #


tzarebczan:
Thanks for the fix!

PR #1322, opened 2020-07-02
Community servers for Hardware wallets

kodxana:
Added Community servers in Troubleshooting section so users wont be off when LBRY SPV1 have problems and they can pick up one of the community server instead.


dheniel23:
What is that ??

PR #1324, opened 2020-07-11
Create services-monitor.md

eatdostacos:
### Description

I created a simple page explaining what some of the services show on the service monitor do. This is in conjunction with my pull request on the LBRY-Desktop client adding a link to this page and the services monitor in the help page.

### Fixes #

### DIFF
--- a/content/faq/services-monitor.md
+++ b/content/faq/services-monitor.md
@@ -1,0 +1,44 @@
+---
+title: LBRY Services Monitoring
+category: troubleshooting
+---
+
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for the PR! I guess you'd need us to fill in some of the details for the other services? 

Will circle back with the team to make sure we want to use this page going forward. 

@nikooo777 we'd probably want to add a few more things like players/lbry.tv apis to status if we plan to use this page.

kauffj:
we've decided we don't want to be pushing the current status page, it's too confusing for a public status page

PR #1331, opened 2020-09-04
Remove extraneous period.

johnzilla:


tzarebczan:
Thanks @johnzilla ! Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

johnzilla:
Sure, but not needed. :) Just saw it and needed an excuse to actually do my first-ever pull request vs just messing around on my laptop.

PR #1338, opened 2020-09-29
Introduce a QR code for a more user-friendly APK download

danger89:
Introduce an easy way to install the Android APK from your web browser on your mobile file, using a QR code.

Almost all Android phones have a QR scan app (even built-in the camera app), which allow you to easily scan the QR code. Which will bring you to the latest APK version (QR code goes to URL: http://lbry.com/releases/lbry-android.apk)

Especially now when Google blocked LBRY in the Android Play Store. But also later, when its back. Its much easier for the end-user to download the APK file this way.

**Edit**: Google unblocked LBRY until further notice in Android Play Store. 

Thanks!

Regards,
Melroy


danger89:
Thanks Jeremy for picking up this pr yourself 👍.

kauffj:
@danger89 is this actually a popular/desirable feature? I checked a few other mobile-focused companies and they appear to not use a QR code on their download pages.

danger89:
> @danger89 is this actually a popular/desirable feature? I checked a few other mobile-focused companies and they appear to not use a QR code on their download pages.

I think it could be useful for people who are not able to download the app from the Play Store (for several reasons; you know by now). Especially if lbry is censored, this is a great alternative.

kauffj:
@danger89 agreed on providing direct APK downloads, which are already linked on the page. Can you share any example of another company/brand that provides direct APK download links also providing a QR code? It just seems like a lot of page real estate for an uncommon download path (loading lbry.com/android on your desktop computer).

danger89:
I don't know. If you just don't like it, reject the pr.

PR #1342, opened 2020-10-07
Update exchanges.md

rywards:
The 3 of the 4 instant exchanges don't list LBC at all. Swapzone has it listed but offers are hard to find.

### Description

3 of the 4 instant exchanges on the page don't even have LBC listed anymore. 

### Fixes #

Removed listings to instant exchanges that are out of date.


tzarebczan:
They are under maintenance per the exchanges and will be back (we hope). Will remove if not back next week.

kauffj:
At least some of these are back

kauffj:
Tom fixed the list

PR #1347, opened 2020-10-23
Added F-Droid links for android.

LaroTayoGaming:
### Description
Just added an F-droid link on lbry.com/get for f-droid users.


tzarebczan:
Nice, thank you and congrats on the PR! Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

PR #1350, opened 2020-10-27
F-Droid Badges, deprecated youtube partner program, and add a redirect for the nav

LaroTayoGaming:
### Description

1. Added Use Odysee on mobile nav. Desktop nav soon if odysee will launch an app or whatever.

2. Added the F-Droid badge for a much detailed approach.

3. Deprecated lbry.com/youtube and add a redirect for the lbry.tv one. (Note for the devs: Add a frontend look for lbry.tv/youtube), and edited a new nav to sync content.

**NOTE: I have made a mistake by deleting youtube.php to the class, but it makes a PHP error. So I need your help. Reverted btw. It is ok now without errors👍🏼**


tzarebczan:
Thanks for the PR! Please keep the redirect stuff in a separate PR - you only want to include one change at a time. We'll get this reviewed once it's backed out.

tzarebczan:
Jeremy is taking care of the /youtube page - can you separate out any other changes into a new PR (each on their own)?

PR #1359, opened 2020-12-31
better and more descriptive links for reporting

cauerego:
also they numbered list there is both wrong and hard to see in android... which in itself could be a bug somewhere! 😘

### Description

it's a minor improvement, from my point of view, but might help people to get better directions at how to proceed with reporting issues with github.

### DIFF
--- a/content/faq/support.md
+++ b/content/faq/support.md
@@ -17,8 +17,8 @@ If you're using LBRY Desktop, LBRY log files will help us better understand the 
 
 To report an issue, you can do one of the following:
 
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for the PR! Can we show you some [appreciation](lbry.com/faq/appreciation)?

cauerego:
@tzarebczan you probably meant an "appreciation with http":

https://lbry.com/faq/appreciation

which is absolutely lovely, thanks!

PR #1361, opened 2021-01-20
Clarifying what kind of effect blocking has

keikari:
### Description

Clearly state that blocking channel only affects what content the one who blocked can see. 
I thought it would be good to have some black on white of how it works, as some think that blocking prevents owner of blocked channel from accessing to your channel.
(There was question about this in discord server, and info provided was bit mixed. I tested the blocking feature myself and it appeared to work as I've added to file.)

How I tested:
1. Created virtual machine running Ubuntu and download lbry AppImage
2. Launched lbry on virtual machine and send few LBCs to that wallet so channel could be created.
3. On lbry running on host machine, I blocked newly created channel. And waited like 15min just to be sure that if something happens it has some time to do that.
4. Back to virtual machine, Tried to search videos which are tied to wallet on host machine and was able to access content normally

### DIFF
--- a/content/faq/lbry-basics.md
+++ b/content/faq/lbry-basics.md
@@ -103,7 +103,7 @@ You can subscribe to a channel by clicking the Follow button from the homepage, 
 ![Manage tags](https://spee.ch/@lbryfaqsc:7/LBRY-manage-tags.jpg)
 
 ## How can I block channels?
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for this, appreciate it!

PR #1373, opened 2021-03-13
Update benefit-consumers.md - "paid to seed" phrase misleading

klesun:
Make it more clear that _paid to seed_ feature is not implemented yet by linking the phrase to the [host-content](https://lbry.com/faq/host-content) page which has following bold warning in the bottom:

> Please note: Hosting fees are currently disabled on the network as we're currently working on a better way to implement the fee structure.

(was the big reason I was into LBRY, resulted in a big downer)

### Fixes #


klesun:
I don't insist on this specific change, just wanted to point out that there is a potential problem in current wording with engaged contributos.

tzarebczan:
Thank you for the edit and effort to clear this up, much appreciated!

PR #1374, opened 2021-03-19
naming.md: Fix claimID/username syntax

jivanpal:
### Description

[The spec](https://lbry.tech/spec#claim-id) says `#` syntax is deprecated. Also, minor clarifying change, writing `@<channel>` rather than `<@channel>`.

### Fixes #


tzarebczan:
Thank you! Both # and : will work for the time being. 

Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

PR #1375, opened 2021-03-23
Clear confusion about `Download directory` vs blobfiles in lbry-directories.md

klesun:
### Description

See https://github.com/lbryio/lbry-desktop/issues/3891

Me, and I believe many other people mistakenly assumed the `Download directory` in the settings to be the place where seeded content is stored. There are some tickets of people getting surprised after getting notification about no space left on system disk despite them specifying this directory to be on a different volume.

After adding this note I hope it will become easier for people to understand what are the steps to seed properly and not worry about their system disk getting clogged.

Feedback and corrections to the message are also welcome.

### Fixes https://github.com/lbryio/lbry-desktop/issues/3891


tzarebczan:
Thank you! 

Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

klesun:
Sure, if you won't have more problems with SEC because of that =-D

Thanks for accepting the PR!

PR #1377, opened 2021-04-03
Update the "Daily Views" Description values

afoxtier:
### Description

Updated description of "Daily Views" to reflect current values. Note: I don't know if the "0.22" in the reward section is still correct and is just rounded in Odysee.


tzarebczan:
Thank you, fixed the .22 also!

PR #1378, opened 2021-04-05
Update location of the help menu

danger89:
### Description

I updated and improved the text on where to find the Help menu in the LBRY Desktop App. As well as where to find the wallet on disk.

This avoid confusing, since the documentation was not inline with the latest version. And lacking some details as well.

Regards,
Melroy van den Berg

Address: bWAxMn8dbKzDfnXrU33rtrByMoZQa1FGjw


tzarebczan:
thank you!

PR #1385, opened 2021-04-29
PHP 8 Compatibility update for lbry.com

iamflorencejay:
### Description

Updated README.md, INSTALL.md, and updated composer to work on PHP 8


lyoshenka (changes requested):
there's a lot going on here. i'm not clear on the purpose or the logic behind it. can you please explain

- title says php 8 but you're requiring php 7.2+
- composer install instructions are not needed. we have `composer.phar` checked into the repo
- why the tagline change?
- why the security.md file? we don't have specific php versions that we do or do not support.
- whats Roave Security Advisories? whats tosdr.org?

iamflorencejay:
@lyoshenka
1. This update is to add the compatibility for PHP 8. So it will run lbry.com smoothly on PHP 8.
2. okay. edit that out.
3. what tagline?
4. Security.md file is based on the security tab on Github. and it has a warning on it, obviously.
5. Roave Security Advisories is a notifier if a composer plugins have flaws, deprecation, etc. TOSDR on the other hand is like TL;DR but it is based on terms of service. It is just a badge though

lyoshenka:
gotcha. this was confusing because the PR is titled "add php8 compat" but it actually includes two other separate things: the security.md file and the roave dependency.

the php8 fix and the roave dependency are helpful. the security.md file is not something we want for this repo. i understand that github recommends it, but it's not right for us. it doesn't add much and creates complexity we dont want (an extra place to look for information).

could you please squash your commits into one and remove the security.md-related changes? then i'll be happy to merge this. we're also happy to show you some [appreciation](https://lbry.com/faq/appreciation) for your work :smile:

iamflorencejay:
righty-o!

iamflorencejay:
@lyoshenka Done!

lyoshenka:
:+1: 

merged in d549fb4c74933917eb4b87a96f6c4c9496c4e41c

PR #1388, opened 2021-04-30
Help LBRY Save Crypto update

iamflorencejay:
### This is help lbry save crypto update version 2: electric boogaloo.

Added help lbry save crypto and as featured pages


iamflorencejay:
for some reason, it didn't copied the featured folder. I will do it tommorow.

PR #1391, opened 2021-05-08
support: Fix visual glitch in unordered list

einarf:
Newlines between each list entry creates glitched unordered list in the web version.

### Description

Having newlines between each list entry makes a glitched unordered list (all browsers + mobile) on the webpage: https://lbry.com/faq/support

bilde

This document was last addressed in #1359 were proper links was added, but the glitched list remained (even when changed from numbered list). Having no newlines between entries are also in line with other documents using `-` for lists and displaying them correctly (even list entries spanning multiple lines)

I saw this when getting a message on Odysee: "This account must undergo review before you can participate in the rewards program". I'll pass the review, no worries 😄


tzarebczan:
@einarf can we show you some [appreciation](https://lbry.com/faq/appreciation) for this fix?

PR #1393, opened 2021-05-11
Added info about how to change ports

keikari:
Add how to change port + which values for ports will work

### Description

Added info about how to change default ports and which port combinations will work for hosting.

### Fixes #


kauffj:
thanks @keikari, do you have a channel on LBRY?

PR #1395, opened 2021-05-13
The Help LBRY save crypto update

iamflorencejay:
### Description

Added "as featured on techcrunch, etc." on the bottom, help lbry save crypto banner, and updated some stylesheet


kauffj:
we're not going to use the "Help LBRY Save Crypto" header and it also includes an update to PHP 8

if you want to submit a PR with _just_ the "as featured on" update, we'd accept that

iamflorencejay:
copy. @kauffj I didn't see that

PR #1414, opened 2021-06-06
Updated composer to composer 2.0

iamflorencejay:
### Description

- Updated the whole composer package to composer 2.0, and regular package updates.

Benefits of updating will discuss in their blog post: https://blog.packagist.com/composer-2-0-is-now-available/

- Add github workflows to make a pull request if composer plugins has an update


iamflorencejay:
@tzarebczan idk how they update it using CI/CD, but I will try soon.

PR #1421, opened 2021-06-18
fix typo

jimt:
### Fixes #


tzarebczan:
Thanks so much for the fix!

PR #1425, opened 2021-06-28
Update lbry-basics.md

keikari:
Updated images(used html  tag instead of the markdown syntax, to be able to scale images, hoping it doesn't break anything)
Changed "lbry.tv" to "odysee.com"
Some other minor updates to match current app.


kauffj:
can you keep the image tags as markdown rather than HTML tags?

keikari:
> can you keep the image tags as markdown rather than HTML tags?

It's on markdown now

PR #1440, opened 2021-07-09
FAQ: this is the crux of LBRY's defence against the SEC

tomachinz:
the mint / burn thing. or are they only minted?

### Fixes #


tzarebczan:
This part is covered in this FAQ: https://lbry.com/faq/block-rewards and the others are covered by other FAQs as well, please take a look at lbry.com/faq

tomachinz:
> This part is covered in this FAQ: https://lbry.com/faq/block-rewards and the others are covered by other FAQs as well, please take a look at lbry.com/faq

The block-rewards page is nice, but it does not seem to explain anything about video content or how the bandwidth is paid for. Is it hosted via http or https for instance? ipfs? Actually the word "video" nor "play" or "view" is on that page so it's a battle to understand the protocol. Do you have a CDN?

tzarebczan:
That's covered under https://lbry.com/faq/host-content - we use our own edge instances across the world to create a CDN like network.

PR #1454, opened 2021-09-09
Update privacy-and-data.md

dpintof:
change lbry.tv to odysee.com


tzarebczan:
For now, this is a lbry.tv centric repo and we'll probably be moving these FAQs to Odysee directly where they apply.

PR #1455, opened 2021-09-10
Update odysee.com on mobile header

iamflorencejay:
### Description

Just updating it from lbry.tv to odysee.com


iamflorencejay:
#1458

PR #1461, opened 2021-10-09
update twitter link

ykris45:
lbry io is to old, the new is LBRY COM


tzarebczan:
Thanks Chris!

ykris45:
> Thanks Chris!

I ve emailed hello LBRY a month ago but didnt get reply from LBRY inc Mr Thomas. Can you check it now sir. Thanks and your welcome sir

PR #1467, opened 2021-11-12
Update lbry.tv links in FAQ to odysee.com

lonyelon:
### Description

The FAQ page "what-is-lbry" contained old links to lbry.tv. They have been replaced by links to odysee.com.

PR #1468, opened 2021-11-17
Update standalone-wallet.md

FlorianWendelborn:
### Description

"lbry.tv" just redirects to Odysee now. "Simply" was removed to improve wording


FlorianWendelborn:
Similar to #1467

PR #1475, opened 2021-12-27
Fixed a link because it was bothering me

Shroom2020:
### Description

A link was bothering me. I fixed it.

### Fixes #
There was a link formatted incorrectly. It is now not formatted incorrectly.


lyoshenka:
thanks for the fix @Shroom2020, and sorry for the slow merge. in-app conversion is off right now so i'll close this PR, but when we re-enable it i'll make sure the link formatting is right

PR #1477, opened 2022-01-02
Minor updates to changing email info

keikari:
Added some minor things that users have asked about
- if already logged in, use sync from settings
- added the email address to use


lyoshenka:
:heart:

PR #1485, opened 2022-04-17
Update host-content.md

CSmith31:
### Description
removed lbry.tv reference in first paragraph, also changed Web apps to web app, removed lbry.tv reference, and updated tenses in selfsame web app section.

### Fixes #


lyoshenka:
thanks for the fix @CSmith31

